PR #27, opened 2023-05-16
Debian base for Docker images and build for arm64

tuler:
- Docker base image changed from Ubuntu to Debian
- Publishing of Docker tagged images to both DockerHub and GHCR
- Build images for arm64 using https://depot.dev
- Using repository_owner as the name of docker organization (better in case of forks)


vfusco:
Similar work was done on #113

PR #39, opened 2023-05-22
Add PLIC and VirtIO devices

edubart:
## Tasks

- [x] PLIC is a necessary device we need to implement before VirtIO , its main purpose is to support external interrupts.
- [x] VirtIO console device is a console device to replace the current HTIF console only when interactive mode is enabled.
- [x] VirtIO p9fs is the Plan 9 virtual filesystem, it allows sharing host folders with the guest.
- [x] VirtIO net allows sharing host network with the guest.

## This PR depends on
- https://github.com/cartesi/grpc-interfaces/pull/11
- https://github.com/cartesi/machine-emulator-defines/pull/1
- https://github.com/cartesi/machine-emulator-rom/pull/7
- https://github.com/cartesi/image-kernel/pull/11
- https://github.com/cartesi/machine-tests/pull/9

## Testing

You can test the console, network support and shared directory in one go with the following command:

```
$ cartesi-machine \
    --append-rom-bootargs="single=yes" \
    --net-dev=user \
    --share-dir=/tmp \
    -i "ip link set dev eth0 up && ip addr add 10.0.2.1/24 dev eth0 && ip route add default via 10.0.2.2 dev eth0 && echo 'nameserver 10.0.2.3' > /etc/resolv.conf && mount -t 9p vfs0 /mnt && date -s @`date +%s` >> /dev/null && sh"
Running in interactive mode!

         .
        / \
      /    \
\---/---\  /----\
 \       X       \
  \----/  \---/---\
       \    / CARTESI
        \ /   MACHINE
         '

~ # ping -c 5 cartesi.io
PING cartesi.io (18.67.136.23): 56 data bytes
64 bytes from 18.67.136.23: seq=0 ttl=255 time=13.632 ms
64 bytes from 18.67.136.23: seq=1 ttl=255 time=13.632 ms
64 bytes from 18.67.136.23: seq=2 ttl=255 time=14.784 ms
64 bytes from 18.67.136.23: seq=3 ttl=255 time=14.336 ms
64 bytes from 18.67.136.23: seq=4 ttl=255 time=15.168 ms

--- cartesi.io ping statistics ---
5 packets transmitted, 5 packets received, 0% packet loss
round-trip min/avg/max = 13.632/14.310/15.168 ms
~ # ls /mnt
systemd-private-609fcd8cb5b040df8688a84138a28c47-systemd-logind.service-DzYUso
systemd-private-609fcd8cb5b040df8688a84138a28c47-systemd-resolved.service-kOvEdu
systemd-private-609fcd8cb5b040df8688a84138a28c47-systemd-timesyncd.service-OMZsPY
systemd-private-609fcd8cb5b040df8688a84138a28c47-upower.service-Wl2jTn
~ # stty size
106 125
~ # date
Thu Apr 27 11:23:12 UTC 2023
```

## Implementation references

- The PLIC implementation is based on Diego's old PLIC branch:
https://github.com/cartesi-corp/machine-emulator/tree/feature/plic
With some additions taken while reading the PLIC spec:
https://github.com/riscv/riscv-plic-spec/releases/download/1.0.0_rc6/riscv-plic-1.0.0_rc6.pdf

- The VirtIO generic interface is implemented based on its v1.2 spec:
https://docs.oasis-open.org/virtio/virtio/v1.2/virtio-v1.2.pdf
These two blog posts from Redhat that explains how VirtIO and Virtqueue works as quick introduction to the topic:
https://www.redhat.com/en/blog/virtio-devices-and-drivers-overview-headjack-and-phone
https://www.redhat.com/en/blog/virtqueues-and-virtio-ring-how-data-travels

- The Plan9 virtual filesystem is based on the "9P2000.L" protocol spec:
https://github.com/chaos/diod/blob/master/protocol.md
https://9p.io/magic/man2html/5/0intro

- The Console and NET device is based on VirtIO v1.2 spec.

## Implementation strategy

I tried to come up with an implementation more simple than TinyEMU and less generic than QEMU, only the necessary features we need and actually present in the spec are implemented. The implementation also try to not perform allocations.

I commented a lot on the code, most comments are based on the spec and on my own insights.

## Implementation details (when reviewing)

- HTIF console is always used when virtio console is not enabled
- `--virtio-console` options wad added to enable virtio console, it also replaces `console=hvc0` with `console=hvc1` in bootargs
- `earlycon=sbi` was added to bootargs to enable HTIF console before the kernel actually initialize the real console device, this is necessary to print any kernel message or error before VirtIO console is initialized
- `-i` now implies both `--htif-console-getchar --virtio-console` to make interactive sessions more responsive and have console resize support
- HTIF console still works even for interactive sessions, but now you have to use `cartesi-machine --htif-console-getchar sh`
- `--share-dir=<directory>` option was added to share a host directory
- `--net-dev=user` option was added to enable user-space networking
- VirtIO devices can only be used when `--htif-console-getchar` is enabled, meaning that interactive mode is enabled, otherwise a runtime error is raised. This means that `--virtio-console` depends on `--htif-console-getchar` to work.
- VirtIO devices for now are only intended to be used while prototyping
- VirtIO device PMA peek is always returning false, meaning merkle tree update will always fail when a VirtIO device exists, this is intentional since they are not verifiable for this version
- VirtIO configs were added for C/C++/Lua APIs, however not to gRPC, JSON-RPC or protobuf, because VirtIO devices state cannot be serialized since it depends on host state
- The machine can have a maximum of 16 VirtIO devices
- Interactive sessions with option `-i` by default now uses VirtIO console instead of HTIF console
- Interactive sessions with option `-i` cannot have its machine state stored or root hash computed anymore
- PLIC CSRs additions changed uarch CSR addresses, this will impact machine-solidity-step contract

## Possible enhancements for future releases
- Make generic VirtIO device interface verifiable, implementing  PMA peek for it and then make VirtIO console device verifiable
- Add VirtIO configs to gRPC, JSON-RPC, protobuf
- Improve init system to automatically configure networking, mount shared directories and change wall clock
- Add VirtIO tests, no tests were added so far, everything has been tested manually, I am thinking into resuming working on Lua tests cleanups before adding any new VirtIO tests using Lua
- Add more devices (VirtIO GPU)

I will create issues about these enhancements after this gets merged and more discussion. I don't want to make all of those into this PR, otherwise this work will take way more weeks and will be bigger and difficult to review, so I want to break any VirtIO enhancements in minor improvements for the next releases. This way when can already have feedback from people using it earlier, and also more testing earlier.


src/machine-config.h
```diff
@@ -117,6 +119,33 @@ struct htif_config final {
     bool yield_automatic{false};      ///< Make yield automatic available?
 };
 
+/// \brief VirtIO console device state config
+struct virtio_console_config final {};
+
+/// \brief VirtIO Plan 9 filesystem device state config
+struct virtio_p9fs_config final {
```
diegonehab, line 126: 9p?

diegonehab: Have you looked at virtiofs? Is it simpler/harder, slower/faster?

diegonehab: I would use "host_directory" instead of "shared_path".

diegonehab: And just "tag" for the mount_tag, since this is what the virtio specification uses.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+
+  --virtio-console
+    add a VirtIO console device,
+    this console is more responsive than HTIF console and has support for terminal size
+
+  -i
+    run in an interactive mode using a VirtIO console device
+    same as --htif-console-getchar --virtio-console
+
+  --share-dir=<directory>
```
diegonehab, line 271: Here I would try to use something as similar to Docker as possible. Perhaps even a -v/--volume options. Just so people don't have to learn something new. This means that I would always automount using the specified target path.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
   --auto-reset-uarch-state
     automatically reset state after miroarchitecture halt
 
-  -i or --htif-console-getchar
-    run in interactive mode.
+  --htif-console-getchar
```
diegonehab, line 260: I think this option needs a better help message. 
It should be mutually exclusive with --virtio-console, right? What does it mean to use both? --htif-console-getchar needed for interactivity? (I.e., to use the use that trick that syncs with wall-clock time?) If so, I think we need to separate these concepts.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+    run in interactive mode
+
+  --virtio-console
+    add a VirtIO console device,
+    this console is more responsive than HTIF console and has support for terminal size
+
+  -i
+    run in an interactive mode using a VirtIO console device
```
diegonehab, line 268: See discussion on --htif-console-getchar

src/cartesi-machine.lua
diegonehab: You could add a --virtio-9p option that only sets the tag and the host-directory and does not do automounting.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+  --net-dev=user
+    Add a VirtIO network device using host user-space networking,
+    this effectively allows to use host network from inside the machine with ease.
+
+    This is a user-mode network,
+    meaning you don't need root privilege or any configuration in the host to use it,
+    in most case it should work out of the box.
+
+    While being of use, this mode has the following limitations:
+      - There is an additional an emulation layer of the TCP/IP stack, so there is an overhead.
```
diegonehab, line 289: Can you fix line breaks to make sure they work in 80 columns? I think there are older lines that are too long as well.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+      - Not all IP protocols are emulated, but TCP and UDP should work.
+      - Host cannot connect to guest TCP ports.
+
+    The implementation uses libslirp TCP/IP emulator library.
+
+    To enable networking, execute the following commands inside the guest with root privilege:
+      ip link set dev eth0 up
```
diegonehab, line 296: I would also automatically add the required commands to the initialization. I think you haven't rebased yet, and this is why this isn't done, right?

diegonehab: Maybe even remove this information from the help.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+
+    This argument can be specified multiple times, each filesystem device will have
+    a tag (e.g. vfs0, vfs1, ...) that keeps incrementing to be used with the mount command.
+
+    The filesystem can be mounted in the guest with the following command:
+      mount -t 9p vfs0 /mnt
+
+  --net-dev=user
+    Add a VirtIO network device using host user-space networking,
+    this effectively allows to use host network from inside the machine with ease.
```
diegonehab, line 282: Break sentence.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+    The network settings configuration is fixed to the following:
+      Network:      10.0.2.0
+      Netmask:      255.255.255.0
+      Host/Gateway: 10.0.2.2
+      DHCP Start:   10.0.2.15
+      Nameserver:   10.0.2.3
+
+  --net-dev=<iface>
+    Add a VirtIO network device using host TUN/TAP interface,
+    this effectively allows to use host network from inside the machine.
```
diegonehab, line 310: Break sentence.

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+      ip link set dev br0 up
+      ip link set dev tap0 up
+      ip addr add 192.168.3.1/24 dev br0
+      sysctl -w net.ipv4.ip_forward=1
+      iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+
+    In the example above the host public internet interface is eth0,
+    but this depends in the host.
+
+    Finally start the machine with using --net-dev=tap0 and
```
diegonehab, line 328: I would add these commands to the initialization too (and perhaps remove from the documentation here)

src/cartesi-machine.lua
```diff
@@ -257,8 +257,80 @@ where options are:
...
+  --share-dir=<directory>
+    Add a VirtIO Plan 9 filesystem device sharing a host directory with the guest.
+
+    This argument can be specified multiple times, each filesystem device will have
+    a tag (e.g. vfs0, vfs1, ...) that keeps incrementing to be used with the mount command.
+
+    The filesystem can be mounted in the guest with the following command:
+      mount -t 9p vfs0 /mnt
+
+  --net-dev=user
```
diegonehab, line 280: I'd rename to --virtio-network.

src/cartesi-machine.lua
diegonehab: I'd rename to --virtio-network.

diegonehab (commented):
See in-line comments.

src/machine-c-api.h
```diff
@@ -241,6 +241,54 @@ typedef struct {          // NOLINT(modernize-use-using)
...
+    cm_virtio_console_config console;       ///< Console
+    cm_virtio_p9fs_config p9fs;             ///< Plan 9 filesystem
+    cm_virtio_net_user_config net_user;     ///< User-mode networking
+    cm_virtio_net_tuntap_config net_tuntap; ///< TUN/TAP networking
+} cm_virtio_device_config_union;
+
+/// \brief VirtIO device state configuration
+typedef struct {
+    CM_VIRTIO_DEVICE_TYPE type;           ///< VirtIO device type
+    cm_virtio_device_config_union device; ///< VirtIO device config
```
diegonehab, line 283: What is your opinion about using a nameless union here to avoid writing "device" all the time?

src/machine-c-api.h
```diff
@@ -241,6 +241,54 @@ typedef struct {          // NOLINT(modernize-use-using)
...
+} cm_virtio_device_config;
+
+/// \brief VirtIO device configuration array
+typedef struct { // NOLINT(modernize-use-using)
```
diegonehab, line 287: The convention is to use `cm_virtio_config_array` instead of `cm_virtio_configs`

diegonehab (commented):
See more in-line comments.

src/machine-config.h
edubart: > 9p or p9?

Both names are used in multiple project, the correct is 9P, but P9 is usually used in open source projects because in C names can't begin with numbers, so the code uses P9 for variable names, and file names, and class names.

> Have you looked at virtiofs? Is it simpler/harder, slower/faster?
Yes, I have. It is harder, because it requires an external application running in the host, however it is faster. I don't think it make sense for us to support it, because it's a Linux only solution. While 9P can work on any OS.

src/machine-c-api.h
edubart: I am against using nameless union or structs, because there are not friendly with C binding generators. Also it's not part of C99 spec, they were introduced in C11. And I want the C API to be compatible with C99.

src/machine-c-api.h
```diff
@@ -241,6 +241,54 @@ typedef struct {          // NOLINT(modernize-use-using)
     bool yield_automatic; ///< Make yield automatic available?
 } cm_htif_config;
 
+/// \brief VirtIO device type
+typedef enum { // NOLINT(modernize-use-using)
+    CM_VIRTIO_DEVICE_NONE = 0,
```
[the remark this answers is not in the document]
edubart: It can be removed, will do it.

edubart:
Closed in favor of https://github.com/cartesi/machine-emulator/pull/172

PR #67, opened 2023-05-31
Added deb package

### LINKED ISSUE #65: Debian packaging
## Context

The emulator is currently provided as a docker image and as `.tar.gz`.
Dependency management is left to the user, who is required to carefully follow the documentation to get it to work.
Linux distributions have package managers as a much better solution for this problem, we should use them.

## Possible solutions

Package the `machine-emulator` into a `.deb`.

## Definition of Done

- [ ] `apt install ./<package>` installs files and pulls dependencies starting from a clean installation.

### LINKED ISSUE #106: Distribute versioned shared libraries
## Context

Currently an application that was linked to a newer `libcartesi.so` version such as v0.14.0 is able to run on systems with old emulators version such as v0.13.0 and potentially crash because of mismatches in the C ABI interface. This is because the shared library always has the name `cartesi.so`, and there is no option to link to a versioned shared library.

This is also a problem in an environment where different emulator versions coexist in different paths, a typical situation when developing a new emulator while having to test an old emulator.

## Possible solutions

We should start distributing a versioned shared library, such as `libcartesi.so.0.14.0` and make `libcartesi.so` a symlink to it.
This way we can link programs to versioned shared libraries, and then attempts to run an application that was actually linked to a different emulator will give an error about the shared library not found instead of a crash or undefined behavior. We should also make use of this, that is, start to link a versioned library for `cartesi.so`.

mpolitzer:
build (On a debian/ubuntu docker container):
```
# 1. build the emulator and uarch
make -j$(nproc) dep
make -j$(nproc)
make -j$(nproc) uarch

# 2. create the package
make deb DESTDIR=$PWD/_install PREFIX=/usr MACHINE_EMULATOR_VERSION=0.14.0-dev
```

use:
```
# 1. install .deb in a clean image
apt update
apt install ./machine-emulator.deb

cartesi-machine --help
...
```

close #65
close #106


deb-control
```diff
@@ -0,0 +1,11 @@
+Package: machine-emulator
+Source: machine-emulator
+Version: 0.0.0
+Architecture: amd64
```
tuler, line 4: Don't forget to give `arm64` some love ❤️

mpolitzer: It shouldn't be complicated if we add `arm64` to CI. But that is out of this PR scope.
You can check the debian name of your architecture locally by running: `dpkg --print-architecture` and editing the file in the meantime. But I would wait for a bit, this PR is still WIP.

tools/template/cartesi-machine.template
```diff
@@ -2,4 +2,4 @@
 export LUA_PATH="$(lua5.3 -e 'print(package.path)');ARG_LUA_PATH"
```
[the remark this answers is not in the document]
edubart: An option to completely avoid this problem:

If we install things correctly in `/usr`, this proxy script is unnecessary, we could simplify and install `/usr/bin/cartesi-machine.lua` directly as `/usr/bin/cartesi-machine`. The images path should also be auto detectable. I think this proxy script is more relevant when we are developing rather than distributing.

edubart (changes requested):

Makefile
```diff
@@ -2,15 +2,28 @@ UNAME:=$(shell uname)
 
 # Install settings
 PREFIX= /opt/cartesi
-BIN_INSTALL_PATH= $(PREFIX)/bin
-LIB_INSTALL_PATH= $(PREFIX)/lib
-SHARE_INSTALL_PATH= $(PREFIX)/share
+DESTDIR ?= $(PREFIX)
+MACHINE_EMULATOR_VERSION= 0.14.0
```
mpolitzer: I wish we avoid hard coding constants, otherwise we may have to keep updating this every release, what about leaving some unspecified default and let the caller decide this, like 0.0.0, or dev, or empty, or retrieve from git, not sure.

NOTE: recreated from a deletion mistake. (originally by: @edubart)

mpolitzer:
@alexmikhalevich Sure.
`make debian` is creating a binary package. The catch is that it has to be built on the same distro it will run.
Maybe the name is misleading (would `make deb` be better?).

Can you try doing both parts (build and run) in docker?
The first half of `.github/workflows/Dockerfile` can be used to install the build dependencies.

Makefile
```diff
@@ -2,15 +2,28 @@ UNAME:=$(shell uname)
...
+DEB_FILENAME= machine-emulator-v$(MACHINE_EMULATOR_VERSION).deb
+BIN_INSTALL_PATH= $(DESTDIR)/bin
+BIN_RUNTIME_PATH= $(PREFIX)/bin
+LIB_INSTALL_PATH= $(DESTDIR)/lib
+ifeq ($(PREFIX),/opt/cartesi)
+SHARE_INSTALL_PATH= $(DESTDIR)/share
+SHARE_RUNTIME_PATH= $(PREFIX)/share
+else
+SHARE_INSTALL_PATH= $(DESTDIR)/share/cartesi
+SHARE_RUNTIME_PATH= $(PREFIX)/share
```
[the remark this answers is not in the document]
mpolitzer: I'm not sure about changelog, `debian` expects a specific format[1] that we don't use.
Trying to read it with tools fails like this[2], didn't investigate further.

Also, `apt install` skips unexpected files.
tried adding `CHANGELOG.md.gz` and `README.md.gz` to the `.deb` (with and without compressing).
But they don't get installed.

[1] https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog
[2]
```
$ apt-get changelog cartesi-machine
E: Failed to fetch changelog:/cartesi-machine.changelog  Changelog unavailable for cartesi-machine=0.0.0
```

tools/template/cartesi-machine.template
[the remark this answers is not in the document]
mpolitzer: Should be fixed on the fixups.

Makefile
```diff
@@ -243,7 +265,12 @@ install-strip:
 	cd $(LIB_INSTALL_PATH) && $(STRIP_EXEC) $(DEP_TO_LIB)
 	cd $(LUA_INSTALL_CPATH) && $(STRIP_EXEC) *.so
 
-install: install-dep install-emulator install-strip $(INSTALL_PLAT)
+install: install-dep install-emulator install-strip $(INSTALL_PLAT) install-tests
+debian: install-dep install-emulator install-strip $(INSTALL_PLAT)
```
edubart, line 269: I agree with the suggestion about renaming to `make deb` or `make deb-package`. So we can repurpose this to make packages for any .deb based distro (Ubuntu, Debian or other).

.github/workflows/build.yml
```diff
@@ -54,6 +54,9 @@ jobs:
       - name: Build uarch-based interpreter
         run: make uarch-with-toolchain
 
+      - name: Build deb package
+        run: make debian DESTDIR=$PWD/_install/usr PREFIX=/usr DEB_FILENAME=machine-emulator.deb
```
edubart, line 58: I think you should define `MACHINE_EMULATOR_VERSION` version here, at least on tag releases to the deb package is built correctly.

Makefile
```diff
@@ -2,15 +2,29 @@ UNAME:=$(shell uname)
 
 # Install settings
 PREFIX= /opt/cartesi
-BIN_INSTALL_PATH= $(PREFIX)/bin
-LIB_INSTALL_PATH= $(PREFIX)/lib
-SHARE_INSTALL_PATH= $(PREFIX)/share
+DESTDIR ?= $(PREFIX)
```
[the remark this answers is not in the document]
mpolitzer: Changing to `$(DESTDIR)$(PREFIX)` on most places, should give the expected behavior.

Makefile
```diff
@@ -260,8 +262,8 @@ install-strip:
 	cd $(LIB_INSTALL_PATH) && $(STRIP_EXEC) $(DEP_TO_LIB)
 	cd $(LUA_INSTALL_CPATH) && $(STRIP_EXEC) *.so
 
-install: install-dep install-emulator install-strip $(INSTALL_PLAT) install-tests
-deb: install-dep install-emulator install-strip $(INSTALL_PLAT)
+install: install-dep install-emulator install-strip $(INSTALL_PLAT)
+deb: install
 	mkdir -p $(DESTDIR)/DEBIAN $(DESTDIR)/share/doc/cartesi-machine/
```
edubart, line 267: The prefix is missing for statements in this target after the recent changes. Maybe use `SHARE_INSTALL_PATH` ?

tools/template/cartesi-machine-stored-hash.template
```diff
@@ -1,2 +1,5 @@
 #!/bin/sh
-LUA_PATH="${LUA_PATH};ARG_LUA_PATH" LUA_CPATH="${LUA_CPATH};ARG_LUA_CPATH" lua5.3 ARG_BIN_INSTALL_PATH//cartesi-machine-stored-hash.lua "$@"
+export LUA_PATH="$(lua5.3 -e 'print(package.path)');ARG_LUA_PATH"
+export LUA_CPATH="$(lua5.3 -e 'print(package.cpath)');ARG_LUA_CPATH"
+export CARTESI_IMAGES_PATH="ARG_INSTALL_PATH"
```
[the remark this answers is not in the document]
mpolitzer: fixed.

Makefile
```diff
@@ -203,38 +219,44 @@ uarch-with-toolchain:
...
 		done
 
 install-Linux:
-	cd $(BIN_INSTALL_PATH) && for x in $(DEP_TO_BIN) $(EMU_TO_BIN); do patchelf --set-rpath $(LIB_INSTALL_PATH) $$x ; done
-	cd $(LIB_INSTALL_PATH) && for x in `find . -maxdepth 1 -type f -name "*.so*"`; do patchelf --set-rpath $(LIB_INSTALL_PATH) $$x ; done
-	cd $(LUA_INSTALL_CPATH) && for x in `find . -maxdepth 2 -type f -name "*.so"`; do patchelf --set-rpath $(LIB_INSTALL_PATH) $$x ; done
+	cd $(BIN_INSTALL_PATH) && for x in $(DEP_TO_BIN) $(EMU_TO_BIN); do patchelf --set-rpath $(LIB_RUNTIME_PATH) $$x ; done
+	cd $(LIB_INSTALL_PATH) && for x in `find . -maxdepth 1 -type f -name "*.so*"`; do patchelf --set-rpath $(LIB_RUNTIME_PATH) $$x ; done
+	cd $(LUA_INSTALL_CPATH) && for x in `find . -maxdepth 2 -type f -name "*.so"`; do patchelf --set-rpath $(LIB_RUNTIME_PATH) $$x ; done
+	patchelf --set-rpath $(LIB_RUNTIME_PATH) src/tests/test-machine-c-api
```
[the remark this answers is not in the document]
mpolitzer: yes, moved.

Makefile
```diff
@@ -243,7 +265,12 @@ install-strip:
 	cd $(LIB_INSTALL_PATH) && $(STRIP_EXEC) $(DEP_TO_LIB)
 	cd $(LUA_INSTALL_CPATH) && $(STRIP_EXEC) *.so
 
-install: install-dep install-emulator install-strip $(INSTALL_PLAT)
+install: install-dep install-emulator install-strip $(INSTALL_PLAT) install-tests
```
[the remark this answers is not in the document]
mpolitzer: fixed.

tools/template/cartesi-machine-stored-hash.template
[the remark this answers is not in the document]
edubart: The line `export CARTESI_IMAGES_PATH="ARG_INSTALL_PATH"` still there for `/usr/bin/cartesi-machine-stored-hash`

Makefile
```diff
@@ -203,38 +216,46 @@ uarch-with-toolchain:
...
-	cat tools/template/cartesi-machine-stored-hash.template | sed 's|ARG_LUA_PATH|${LUA_DEFAULT_PATHS}|g;s|ARG_LUA_CPATH|${LUA_DEFAULT_C_PATHS}|g;s|ARG_BIN_INSTALL_PATH|${BIN_INSTALL_PATH}|g' > $(BIN_INSTALL_PATH)/cartesi-machine-stored-hash
-	cat tools/template/rollup-memory-range.template | sed 's|ARG_LUA_PATH|${LUA_DEFAULT_PATHS}|g;s|ARG_LUA_CPATH|${LUA_DEFAULT_C_PATHS}|g;s|ARG_BIN_INSTALL_PATH|${BIN_INSTALL_PATH}|g' > $(BIN_INSTALL_PATH)/rollup-memory-range
-	cat tools/template/uarch-riscv-tests.template | sed 's|ARG_LUA_PATH|${LUA_DEFAULT_PATHS}|g;s|ARG_LUA_CPATH|${LUA_DEFAULT_C_PATHS}|g;s|ARG_BIN_INSTALL_PATH|${BIN_INSTALL_PATH}|g' > $(BIN_INSTALL_PATH)/uarch-riscv-tests
-	cd $(BIN_INSTALL_PATH) && $(CHMOD_EXEC) $(EMU_TO_BIN) cartesi-machine cartesi-machine-tests cartesi-machine-stored-hash rollup-memory-range
+	cat tools/template/cartesi-machine.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_DEFAULT_C_PATHS)|g;s|ARG_INSTALL_PATH|$(IMAGES_RUNTIME_PATH)|g;s|ARG_BIN_RUNTIME_PATH|$(BIN_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/cartesi-machine
```
edubart, line 253: I packaged as the following:

```
make deb DESTDIR=$PWD/_install PREFIX=/usr/x86_64 MACHINE_EMULATOR_VERSION=0.14.0-dev
```

And this is the contents of `/usr/x86_64/bin/cartesi-machine`:
```
#!/bin/sh
export LUA_PATH="$(lua5.3 -e 'print(package.path)');/usr/x86_64/share/lua/5.3/?.lua"
export LUA_CPATH="$(lua5.3 -e 'print(package.cpath)');./?.so;/usr/local/lib/lua/5.3/?.so;/usr/local/share/lua/5.3/?.so;/opt/cartesi/lib/lua/5.3/?.so"
export CARTESI_IMAGES_PATH="/usr/x86_64/share/cartesi-machine/images"
lua5.3 "/usr/x86_64/bin/cartesi-machine.lua" "$@"
```

`LUA_PATH` is correct, but `LUA_CPATH` does not include the installed files in `/usr/x86_64/` and still references `/opt/cartesi/`. 

I think `LUA_DEFAULT_PATHS` and `LUA_DEFAULT_C_PATHS` are obsolete, and could be removed from the makefile, instead make use of `LUA_RUNTIME_CPATH/?.so`.

mpolitzer: Updated. running: `rm -rf _install/; make deb DESTDIR=$PWD/_install PREFIX=/usr/a/b/c` gave:

```
cat /usr/a/b/c/bin/cartesi-machine
#!/bin/sh
export LUA_PATH="$(lua5.3 -e 'print(package.path)');/usr/a/b/c/share/lua/5.3/?.lua"
export LUA_CPATH="$(lua5.3 -e 'print(package.cpath)');/usr/a/b/c/lib/lua/5.3/?.so"
export CARTESI_IMAGES_PATH="/usr/a/b/c/share/cartesi-machine/images"
lua5.3 "/usr/a/b/c/bin/cartesi-machine.lua" "$@"
```

And runs as:
```
/usr/a/b/c/bin/cartesi-machine
lua5.3: /usr/a/b/c/bin/cartesi-machine.lua:1025: unable to find length of flash drive 'root' image file '/usr/a/b/c/share/cartesi-machine/images/rootfs.ext2'
stack traceback:
        [C]: in function 'assert'
        /usr/a/b/c/bin/cartesi-machine.lua:1025: in local 'resolve_flash_lengths'
        /usr/a/b/c/bin/cartesi-machine.lua:1130: in main chunk
        [C]: in ?
```

edubart: But this error is because this package does not install `rootfs.ext2` right? If you place it at `/usr/a/b/c/share/cartesi-machine/images/` should work I think.

mpolitzer: That was to show `cartesi-machine` running as expected.
Not to indicate a problem.
I looks misleading in retrospect.

edubart (approved):
I approve, good work! With this PR, not only will it be easier to package for Debian-based distributions, but also for any distribution.

Makefile
```diff
@@ -2,13 +2,15 @@ UNAME:=$(shell uname)
...
 PREFIX= /opt/cartesi
-BIN_INSTALL_PATH= $(PREFIX)/bin
-LIB_INSTALL_PATH= $(PREFIX)/lib
-SHARE_INSTALL_PATH= $(PREFIX)/share
-IMAGES_INSTALL_PATH= $(SHARE_INSTALL_PATH)/images
-LUA_INSTALL_CPATH= $(PREFIX)/lib/lua/5.4
-LUA_INSTALL_PATH= $(PREFIX)/share/lua/5.4
-INC_INSTALL_PATH= $(PREFIX)/include/machine-emulator
+MACHINE_EMULATOR_VERSION:= $(shell make -sC src get-version)
+DEB_FILENAME= machine-emulator-v$(MACHINE_EMULATOR_VERSION).deb
```
edubart, line 6: Should the deb name be `cartesi-machine-*` by default, like the package name?

edubart (changes requested):

Makefile
mpolitzer: I like cartesi-machine better as well.
But I changed this from the last review to use the same naming the `.tar.gz`.

Makefile
```diff
@@ -26,11 +37,13 @@ DEP_TO_BIN=
 DEP_TO_LIB=
 EMU_TO_BIN= jsonrpc-remote-cartesi-machine remote-cartesi-machine remote-cartesi-machine-proxy merkle-tree-hash
 EMU_TO_LIB= $(LIBCARTESI_$(UNAME)) $(LIBCARTESI_PROTOBUF_$(UNAME)) $(LIBCARTESI_GRPC_$(UNAME))
-EMU_LUA_TO_BIN= cartesi-machine-tests.lua cartesi-machine.lua cartesi-machine-stored-hash.lua rollup-memory-range.lua uarch-riscv-tests.lua
+EMU_LUA_TEST_TO_BIN= cartesi-machine-tests.lua uarch-riscv-tests.lua
+EMU_LUA_TO_BIN= cartesi-machine.lua cartesi-machine-stored-hash.lua rollup-memory-range.lua
 EMU_TO_LUA_PATH= cartesi/util.lua cartesi/proof.lua
```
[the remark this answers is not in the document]
mpolitzer: added to share.

Makefile
mpolitzer: changed to cartesi-machine.

Makefile
```diff
@@ -201,38 +212,46 @@ uarch-with-toolchain:
...
-	cd $(BIN_INSTALL_PATH) && $(CHMOD_EXEC) $(EMU_TO_BIN) cartesi-machine cartesi-machine-tests cartesi-machine-stored-hash rollup-memory-range
+	cat tools/template/cartesi-machine.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_RUNTIME_CPATH)/?.so|g;s|ARG_INSTALL_PATH|$(IMAGES_RUNTIME_PATH)|g;s|ARG_BIN_RUNTIME_PATH|$(BIN_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/cartesi-machine
+	cat tools/template/cartesi-machine-stored-hash.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_RUNTIME_CPATH)/?.so|g;s|ARG_BIN_RUNTIME_PATH|$(BIN_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/cartesi-machine-stored-hash
+	cat tools/template/rollup-memory-range.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_RUNTIME_CPATH)/?.so|g;s|ARG_BIN_RUNTIME_PATH|$(BIN_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/rollup-memory-range
+	cd $(BIN_INSTALL_PATH) && $(CHMOD_EXEC) $(EMU_TO_BIN) cartesi-machine cartesi-machine-stored-hash rollup-memory-range
 	cd $(BIN_INSTALL_PATH) && $(CHMOD_DATA) $(EMU_LUA_TO_BIN)
```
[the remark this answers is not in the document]
mpolitzer: moved to share.

edubart (dismissed):

edubart (dismissed):

Makefile
```diff
@@ -2,22 +2,42 @@ UNAME:=$(shell uname)
...
-IMAGES_INSTALL_PATH= $(SHARE_INSTALL_PATH)/images
-UARCH_INSTALL_PATH= $(SHARE_INSTALL_PATH)/uarch
-LUA_INSTALL_CPATH= $(PREFIX)/lib/lua/5.4
-LUA_INSTALL_PATH= $(PREFIX)/share/lua/5.4
-INC_INSTALL_PATH= $(PREFIX)/include/machine-emulator
+MACHINE_EMULATOR_VERSION:= $(shell make -sC src get-version)
+MACHINE_EMULATOR_SO_VERSION:= $(shell make -sC src get-so-version)
+DEB_FILENAME= cartesi-machine-v$(MACHINE_EMULATOR_VERSION).deb
+BIN_RUNTIME_PATH= $(PREFIX)/bin
+LIB_RUNTIME_PATH= $(PREFIX)/lib/cartesi-machine
```
edubart, line 9: I noticed in recent changes that we don't need to provide grpc libraries anymore. With the change the `cartesi-machine` folder in `lib` becomes less necessary. What do you think about removing it? So we install `$(PREFIX)/lib/libcartesi.so` instead of `$(PREFIX)/lib/cartesi-machine/libcartesi.so`

tools/template/cartesi-machine.template
```diff
@@ -2,4 +2,4 @@
 export LUA_PATH_5_4="${LUA_PATH_5_4:-;};ARG_LUA_PATH"
```
edubart, line 2: I started to have some issues regarding multiple installed versions of cartesi machine in my environment.
For example I have a cartesi-machine machine installed at `/usr` and another one in `/opt/cartesi`, when I execute `/opt/cartesi/bin/cartesi-machine` the `LUA_PATH_5_4` tries to look up first the installation at `/usr`, instead of the one at `/opt/cartesi`, inverting the order here should fix this problem. That is, place `ARG_LUA_PATH` before:

```bash
export LUA_PATH_5_4="ARG_LUA_PATH;${LUA_PATH_5_4:-;}"
```

And then we should do this for all templates.

edubart (changes requested):

Makefile
```diff
@@ -25,13 +47,14 @@ STRIP_EXEC= strip -x
 DEP_TO_BIN=
 DEP_TO_LIB=
 EMU_TO_BIN= jsonrpc-remote-cartesi-machine remote-cartesi-machine remote-cartesi-machine-proxy merkle-tree-hash
-EMU_TO_LIB= $(LIBCARTESI_$(UNAME)) $(LIBCARTESI_PROTOBUF_$(UNAME)) $(LIBCARTESI_GRPC_$(UNAME))
-EMU_LUA_TO_BIN= cartesi-machine-tests.lua cartesi-machine.lua cartesi-machine-stored-hash.lua rollup-memory-range.lua uarch-riscv-tests.lua
-EMU_TO_LUA_PATH= cartesi/util.lua cartesi/proof.lua
+EMU_TO_LIB= $(LIBCARTESI_SO_$(UNAME)) $(LIBCARTESI_SO_PROTOBUF_$(UNAME)) $(LIBCARTESI_SO_GRPC_$(UNAME))
+EMU_LUA_TEST_TO_BIN= cartesi-machine.lua cartesi-machine-tests.lua uarch-riscv-tests.lua cartesi-machine-stored-hash.lua rollup-memory-range.lua
```
edubart, line 51: The files `cartesi-machine-tests.lua` and `uarch-riscv-tests.lua` are always being installed, even though I did not do `make install-tests`.

deb-control
mpolitzer: `arm64` got included. :)

tools/template/cartesi-machine.template
edubart: I made this in https://github.com/cartesi/machine-emulator/pull/114

Makefile
```diff
@@ -241,44 +275,63 @@ uarch-with-linux-env:
...
+	cat tools/template/cartesi-machine.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_RUNTIME_CPATH)/?.so|g;s|ARG_INSTALL_PATH|$(IMAGES_RUNTIME_PATH)|g;s|ARG_LUA_RUNTIME_PATH|$(LUA_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/cartesi-machine
+	cat tools/template/cartesi-machine-stored-hash.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_RUNTIME_CPATH)/?.so|g;s|ARG_LUA_RUNTIME_PATH|$(LUA_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/cartesi-machine-stored-hash
+	cat tools/template/rollup-memory-range.template | sed 's|ARG_LUA_PATH|$(LUA_RUNTIME_PATH)/?.lua|g;s|ARG_LUA_CPATH|$(LUA_RUNTIME_CPATH)/?.so|g;s|ARG_LUA_RUNTIME_PATH|$(LUA_RUNTIME_PATH)|g' > $(BIN_INSTALL_PATH)/rollup-memory-range
+	cd $(BIN_INSTALL_PATH) && $(CHMOD_EXEC) $(EMU_TO_BIN) cartesi-machine cartesi-machine-stored-hash rollup-memory-range
+	cd $(LIB_INSTALL_PATH) && ln -s $(LIBCARTESI_SO_$(UNAME)) $(LIBCARTESI_$(UNAME))
+	cd $(LIB_INSTALL_PATH) && ln -s $(LIBCARTESI_SO_PROTOBUF_$(UNAME)) $(LIBCARTESI_PROTOBUF_$(UNAME))
+	cd $(LIB_INSTALL_PATH) && ln -s $(LIBCARTESI_SO_GRPC_$(UNAME)) $(LIBCARTESI_GRPC_$(UNAME))
```
edubart, line 315: Usually `make install` should overwrite files (when doing upgrades), but this is failing, the fix is to use `ln -sf` here.

edubart (changes requested):

vfusco (approved):

edubart (approved):

PR #76, opened 2023-06-05
Add fork entrypoint to C, C++, and Lua APIs

diegonehab:
The JSONRPC remote machine emulator includes a `fork` entrypoint that forks the server and returns the address of the new server.

This entrypoint was missing from the C, C++, and Lua APIs that interact with the JSONRPC server.

This pull request adds the missing entrypoints and tests for the functionality.


src/tests/test-jsonrpc-fork-lua
edubart, line 1: This file could have `.lua` extension, like other Lua tests. So we have syntax highlighting in github (when reviewing)  and in our editor.

src/tests/test-jsonrpc-fork-lua
```diff
@@ -0,0 +1,172 @@
...
+-- You should have received a copy of the GNU Lesser General Public License
+-- along with the machine-emulator. If not, see http://www.gnu.org/licenses/.
+--
+--
+-- Note: for grpc machine test to work, remote-cartesi-machine must run on
+-- same computer and remote-cartesi-machine execution path must be provided
+-- Note: for jsongrpc machine test to work, jsonrpc-remote-cartesi-machine must run on
+-- same computer and jsonrpc-remote-cartesi-machine execution path must be provided
+
+local cartesi = require "cartesi"
```
edubart, line 24: This variable is set but unused (caught by luacheck).

src/tests/test-jsonrpc-fork-lua
```diff
@@ -0,0 +1,172 @@
...
+        remote_address = o
+        return true
+    end },
+    { ".*", function(all)
+        error("unrecognized option " .. all)
+    end }
+}
+
+-- Process command line options
+local arguments = {}
```
edubart, line 63: This variable is set but unused (caught by luacheck). Could be removed.

src/tests/test-jsonrpc-fork-lua
```diff
@@ -0,0 +1,172 @@
...
+            if option[2](argument:match(option[1])) then
+                break
+            end
+        end
+    else
+        arguments[#arguments+1] = argument
+    end
+end
+
+local remote = jsonrpc.stub(remote_address)
```
edubart, line 76: This variable is set but unused (caught by luacheck).

src/tests/test-jsonrpc-fork-lua
```diff
@@ -0,0 +1,172 @@
...
+    f:close()
+    return name
+end
+
+local stub = jsonrpc.stub(remote_address)
+local config = stub.machine.get_default_config()
+config.rom.image_filename = tmprom()
+config.ram.length = 1 << 22
+
+local machine = stub.machine(config)
```
edubart, line 168: This variable is set but unused (caught by luacheck). But of course removing this line will break the test.

Feels little odd to instantiate a machine and not use it, but I noticed the machine is retrieved later with `stub.get_machine()`.

edubart (commented):
Nice, I tested and it's working. I made some comments the test file, but just about silly things. Optional to fix them.

src/tests/test-jsonrpc-fork-lua
diegonehab: If I don't initialize the machine here, I can initialize it in the recursive function. It's a bit more awkward, but can do.

src/tests/test-jsonrpc-fork-lua
diegonehab: I'll change the way we do this. The problem is that this .lua file is not supposed to be run automatically like the others are during tests. It is supposed to only be run from within the .sh script. So I will change the names of all lua test files and change the "find" to find only the files that are prefixed with autorun or some  such. Any other ideas how to fix this?

diegonehab: I removed the "find" and have now an "autorun" file listing the lua files that should be run automatically during tests.

edubart (approved):

vfusco (approved):

Issue #72, opened 2023-06-04
Speedup remote machine snapshot loading


edubart:
## Context

When writing a set of tests using remote rolling machines it takes several seconds to load the machine from a snapshot. We should investigate what is causing this and find a way to speedup, because it slows down the prototyping of dapp tests.

## Possible solutions

This is probably due to merkle tree computation check on startup, one possibility would be completely disabling it when doing unit tests.

PR #79, opened 2023-06-08
Add runtime options to skip root hash, version check and suppress putchar

edubart:
This issue implements #72

Using the new `--skip-root-hash-check` option we can run a test suite without the merkle tree root hash checking every machine load, saving several seconds when running dapp test suites. I am using this feature with JSONRPC to speed up Lua unit tests I am creating.

Using the new `--skip-version-check` option we can test old snapshots created by Sunodo. Ideally we should not run snapshots targeting old emulator releases, but for development users like us, who need to test stuff before they are actually released, this check is a pain, because it prevents us from testing old snapshots created by Sunodo with emulators in development.

Using the new `--htif-no-console-putchar` option we can suppress anything written to machine's stdout/stderr. This is useful when stress testing dapps that prints to stdout.

I deliberately left out gRPC interface changes related to the new config, mostly to avoid grpc-interface changes and we plan to deprecate it anyway. Let me know if it would be useful to have these options available in gRPC.

### DIFF
--- a/src/machine.h
+++ b/src/machine.h
@@ -743,6 +743,11 @@ public:
         return m_c;
     }
 
+    /// \brief Returns the machine runtime config.
+    const machine_runtime_config &get_runtime_config(void) const {
+        return m_r;
+    }
+
     /// \brief Replaces a memory range.
     /// \param range Configuration of the new memory range.
     /// \details The machine must contain an existing memory range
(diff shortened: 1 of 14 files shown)


mpernambuco (dismissed):
Looks good to me

edubart:
Rebased and added the new option `--suppress-console-output`.

src/machine.h
```diff
@@ -743,6 +743,11 @@ class machine final {
         return m_c;
     }
 
+    /// \brief Returns copy of runtime config.
```
diegonehab, line 746: Not really a copy. :)

edubart: > Code never lies, comments someties do. 
:)

Fixed, it was due to a copy paste of the above comment.

vfusco (approved):

Issue #51, opened 2023-05-26
Check iflags_Y and iflags_H in uarch step


diegonehab:
## Context

When  the big emulator halts or issues a manual yield, it sets the corresponding flag (`iflags_H` and `iflags_Y`, respectively) and enters a _fixed point_ mode. In other words, starting the big emulator when these flags are set does change the state of the machine in any way, not even `mcycle`, leaving the state hash is unchanged.

The uarch emulator has a `uarch_halt_flag` with a similar effect. Starting the uarch emulator when the flag is set also does not change the state of  the machine, not even `uarch_cycle`, leaving the state hash unchanged. 

However, when the big emulator is halted or yielded, running the small emulator does in fact change the state of the machine. This is because the uarch emulator must run the code of the big emulator until that code checks the halt or yield flag and decides there is nothing else to be done, at which point the uarch emulator can set its own halt flag to enter a fixed point.

This is inconvenient for arbitration because it would require us to run the uarch emulator until it halts after running the big emulator until it halts in order to find the fixed point hash. It would be best that, once the big emulator sets its halt or yield flags, that serves as a fixed point for the uarch emulator as well.  This way, we would never need to run the uarch emulator after the big emulator halts or yields.

## Possible solutions

The uarch emulator should check the `iflags_Y` and `iflags_H` flags of the big emulator just like it checks its own `uarch_halt_flag`.

PR #82, opened 2023-06-18
Do not advance microarchitecture when at a fixed point

mpernambuco:
This change ensures that the microarchitecture does not advance micro cycles if it is at a fixed point.
The Cartesi machine is at a fixed point then one of the following conditions is met.
- the microarchitecture cycle counter is the maximum value
- the microarchitecture is halted (uarch_halt_flag set)
- the machine is halted (iflags.H set)
- the machine is manually yielded (iflags.Y set)

Considering that the uarch halts immediately when either iflags.Y or iflags.H is set, the 'big interpreter' loop was modified to ensure that writing to iflags is the final operation.

From: https://github.com/cartesi/machine-emulator/issues/51


src/tests/htif-yield.lua
```diff
@@ -51,14 +128,51 @@ local yields = {
     { mcycle = 520, data = 27, cmd = YIELD_AUTOMATIC, reason = REASON_TX_REPORT},
 }
 
+local function run_machine_with_uarch(machine)
```
edubart, line 131: No need to be done in this PR, but I think this function deserves to be written in C++ and exposed to a C API and Lua API in the future, so it can be faster, speeding up our CI. Doing a round-trip in Lua and C++ for every step is slow. 
Having an API for this I think it might allow us to simplify other tests/scripts doing the same.

@diegonehab Thoughts?

src/uarch-interpret.cpp
```diff
@@ -14,42 +14,35 @@
...
+            case uarch_step_status::success_and_uarch_halted:
+                [[fallthrough]];
+            case uarch_step_status::uarch_halted:
+                return uarch_interpreter_break_reason::uarch_halted;
+            case uarch_step_status::halted:
+                return uarch_interpreter_break_reason::halted;
+            case uarch_step_status::yielded_manually:
+                return uarch_interpreter_break_reason::yielded_manually;
+            case uarch_step_status::at_max_uarch_cycle:
+                return uarch_interpreter_break_reason::reached_target_cycle;
```
edubart, line 42: This line is not covered according to our coverage report. Can you add a test for it?

src/uarch-state-access.h
```diff
@@ -127,6 +127,10 @@ class uarch_state_access : public i_uarch_state_access<uarch_state_access> {
         m_us.halt_flag = false;
     }
 
+    auto do_read_iflags() {
```
edubart, line 130: Why use `auto` in function declaration? We usually are explicit about function return types, so I would change to `uint64_t`.

src/uarch-interpret.cpp
edubart: Thinking more, converting `at_max_uarch_cycle` to a `reached_target_cycle` feels incorrect, because `cycle` will not be actually reached.

What should we do here? Thrown an exception?

@diegonehab Thoughts?

uarch/uarch-machine-state-access.h
```diff
@@ -411,10 +412,12 @@ class uarch_machine_state_access : public i_state_access<uarch_machine_state_acc
         raw_write_memory(shadow_state_get_csr_abs_addr(shadow_state_csr::iflags), val);
     }
 
-    void do_set_iflags_H(void) {
+    NO_RETURN NO_INLINE void do_set_iflags_H(void) {
         auto old_iflags = read_iflags();
         auto new_iflags = old_iflags | IFLAGS_H_MASK;
         write_iflags(new_iflags);
+        // The micro interpreter will never execute this line because it will halt as soon as iflags.H is set
+        __builtin_unreachable();
```
edubart, line 420: But what happens due to a mistake of us (or the compiler), the code reaches here? I think we should make safe for that situation. `__builtin_unreachable()` will let the compiler do crazy optimizations here. I think we could use `__builtin_trap()` which essentially generates an illegal instruction, and this will cause our interpreter to abort.

In case of RISC-V 64, looks like `__builtin_trap()` generates `ebreak`  instruction:
https://gcc.godbolt.org/z/x3aMn4sq9

If doing this I think we should do the same for all other `__builtin_unreachable()`, that is, in `abort()`  and `interpret_next_mcycle_with_uarch()`.

src/uarch-step.cpp
```diff
@@ -1089,5 +1096,55 @@ static inline void uarchExecuteInsn(UarchState &a, uint32 insn, uint64 pc) {
...
+}
+
+static inline bool getIflagsY(uint64 iflags) {
+    return (iflags & IFLAGS_Y_MASK) != 0;
+}
+
+template <typename UarchState>
+uarch_step_status uarch_step(UarchState &a) {
+    // This must be the first read in order to match the first log access in machine::verify_state_transition
+    auto cycle = readCycle(a);
```
edubart, line 1110: If I am not wrong, we can't use `auto` here, otherwise the solidity Lua generator will break.We should use explicit types.

uarch/uarch-runtime.cpp
```diff
@@ -70,13 +70,12 @@ extern "C" void _putchar(char c) {
...
 
-extern "C" [[noreturn]] void abort(void) {
+extern "C" NO_RETURN void abort(void) {
     volatile char *p = reinterpret_cast<char *>(cartesi::uarch_mmio_address::abort);
     *p = uarch_mmio_abort_value;
     // execution will never reach this point
-    // infinite loop added to silence the compiler
-    for (;;) {
-    }
+    asm("1: j 1b"); // infinite loop
```
edubart, line 77: When coding `asm` statements, you should use `asm volatile`, otherwise the compiler is free to change or optimize our assembly code.

edubart (changes requested):
See comments, I marked @diegonehab to give a second feedback on some of them.

uarch/uarch-machine-state-access.h
diegonehab: Depends what you mean by mistake. I don't know why we would "help" the compiler here by adding `_builtin_unreachable()`. The function is marked as [[noreturn]]. It writes to some "random" memory address through a volatile pointer. The compiler can't know what side effect that has. It really is a checkpoint, right? So it already knows everything it needs to know. I my mind, this is the same as a call to `exit` or something.

uarch/uarch-runtime.cpp
diegonehab: I don't think we need the infinite look, right? What is the side effect we'd like to have with `abort`? Do we want to pass out the information that the microarch emulator halted with an error condition?

uarch/uarch-machine-state-access.h
edubart: >  I don't know why we would "help" the compiler here by adding _builtin_unreachable()

True, `__builtin_unreachable()` feels redundant in a function already marked as `[[noreturn]]`.

uarch/uarch-runtime.cpp
edubart: I also feel that `__builtin_trap()` would make more sense here rather than an infinite loop, but again both are pointless, because the interpreter will halt before reaching it.

uarch/uarch-machine-state-access.h
mpernambuco: __builtin_unreachable prevents the compiler from complaining about [[noreturn]] returning.
The exsisting test coverage catches  reordering of write_iflags() or write_mcycle() because it would change the expected mcycle.

edubart: Ok, I am fine with `__builtin_unreachable()` to silence compiler warnings and given the fact that we have tests that will catch compiler write re-ordering.

edubart (dismissed):

src/uarch-interpret.cpp
```diff
@@ -14,42 +14,36 @@
...
+            case uarch_step_status::success_and_uarch_halted:
+                [[fallthrough]];
+            case uarch_step_status::uarch_halted:
+                return uarch_interpreter_break_reason::uarch_halted;
+            case uarch_step_status::halted:
+                return uarch_interpreter_break_reason::halted;
+            case uarch_step_status::yielded_manually:
+                return uarch_interpreter_break_reason::yielded_manually;
+            case uarch_step_status::cycle_overflow:
+                // Prior condition ensures this case is unreachable. but linter may complain about missing it
```
edubart, line 42: Then, wrap this condition with a LCOV comment, to make coverage test 100% and happy, like:

```c++
            // LCOV_EXCL_START
            case uarch_step_status::cycle_overflow:
                // Prior condition ensures this case is unreachable. but linter may complain about missing it
                return uarch_interpreter_break_reason::reached_target_cycle;
            // LCOV_EXCL_STOP
```

diegonehab (approved):

edubart (approved):

PR #84, opened 2023-06-22
Fix remote machine issues

diegonehab:
A stress-test on the `jsonrpc-remote-cartesi-machine` has uncovered a variety of issues:

* The JSONRPC "id" was not being checked in the server to be a number, string, or null;
* The JSONRPC "id" field used by the client is the integer "0", but it was being rejected when "0.0" was returned, which is arguably the same thing;
* We were using Boost Log, which does not play well with fork();
* The `SIGCHLD` handler was messing with `errno`;
* The installed handlers were not setting `SA_RESTART`, so some syscalls were being interrupted in a way that some of our library-dependencies ware not expecting.

To the extent these issues manifested in the gRPC version of `remote-cartesi-machine`, this PR also fixes them.


edubart:
This fixes issues I was having with the jsonrpc remote machine, let's try to merge that soon.

edubart (approved):

vfusco (approved):

PR #88, opened 2023-06-27
feat: add tlb_get_entry_hot/cold_rel_addr

stskeeps:
We only provide abs_addr currently and rel_addr is useful for zk-arch

### DIFF
--- a/src/shadow-tlb.h
+++ b/src/shadow-tlb.h
@@ -98,15 +98,25 @@ static inline bool tlb_is_hit(uint64_t vaddr_page, uint64_t vaddr) {
 }
 
 template <TLB_entry_type ETYPE>
-static inline uint64_t tlb_get_entry_hot_abs_addr(uint64_t eidx) {
-    return PMA_SHADOW_TLB_START + offsetof(shadow_tlb_state, hot) +
+static inline uint64_t tlb_get_entry_hot_rel_addr(uint64_t eidx) {
+    return offsetof(shadow_tlb_state, hot) +
         (ETYPE * sizeof(std::array<tlb_hot_entry, PMA_TLB_SIZE>)) + (eidx * sizeof(tlb_hot_entry));
 }
 
 template <TLB_entry_type ETYPE>
-static inline uint64_t tlb_get_entry_cold_abs_addr(uint64_t eidx) {
-    return PMA_SHADOW_TLB_START + offsetof(shadow_tlb_state, cold) +
+static inline uint64_t tlb_get_entry_cold_rel_addr(uint64_t eidx) {
+    return offsetof(shadow_tlb_state, cold) +
         (ETYPE * sizeof(std::array<tlb_cold_entry, PMA_TLB_SIZE>)) + (eidx * sizeof(tlb_cold_entry));
+}
+
+template <TLB_entry_type ETYPE>
+static inline uint64_t tlb_get_entry_hot_abs_addr(uint64_t eidx) {
+    return PMA_SHADOW_TLB_START + tlb_get_entry_hot_rel_addr<ETYPE>(eidx);
+}
+
+template <TLB_entry_type ETYPE>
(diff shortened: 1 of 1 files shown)


stskeeps:
> thank you Carsten!
> 
> could you please clarify the context of the problem you are solving? is there any related issue opened currently in the repo?

No current issue, but making a long story short:

I'm in my code mapping in shadow TLB to a non-PMA_SHADOW_TLB_START location in memory, so I need the relative offset of an entry and not an absolute one, for my variant of uarch-state-access.h

diegonehab:
Would you be happy if instead we had a function to map virtual to physical addresses using the current mapping? Is this the missing piece you need for messing with the TLB?

stskeeps:
Basically, I am pretty much using the TLB from https://github.com/cartesi/machine-emulator/blob/main/uarch/uarch-machine-state-access.h#L617 as-is, except that my accesses into the shadow TLB is relative, not absolute

Example:
https://github.com/stskeeps/machine-emulator/blob/sts/experimental/risczero-uarch/zkarch/uarch-machine-state-access.h#L659

where I access the hot/cold entry through the relative offset to a location that I've mapped the shadow TLB into. We don't have such an issue in the ordinary machine (as shadow TLB is "generated" from machine state), nor in uarch as uarch sits in same memory space as the shadow TLB is in

I'm in a bit of a third mode where we don't have a separate machine state, only the shadow TLB page(s); which I map in and then need to access in a relative offset manner to where I've mapped it in

diegonehab:
I'm trying to get at the reason why you'd look into the TLB and see if there is a safer way to solve the underlying problem. For context here for the other readers. 

You are trying to run the "interpret" code of the big emulator using risc0, right? And you are writing your own `state_access` object that you will use when running it. As any `state_access` object, it has methods for accessing CSRs, normal registers, but also the TLB and even regular memory. Ostensibly, all these data live in the address space of the big machine. Whenever one of these accesses happens, you need to make sure risc0 has the underlying span of memory where the desired data lives.

In the case of simple registers, for example, they live in the shadow state. If you got the entire shadow state from the machine and made it available to the risc0 machine, you can access it if you convert (within the, say, `risc0_state_access` object) any access to a register to an access to that copy of the shadow state that  was mapped to the risc0 space. You do this by specifying that the access is into the shadow state range, and the offset of the desired register within that range is such and such. 

You can do the same thing for a regular memory access, by using the physical page where the memory item lives.

Now comes an access to the TLB. When interpret accesses a TLB entry via your risc0_state_access, you want to bring the entire tlb shadow into risc0 awareness, and then say that the access is relative to that tlb chunk. But for that, you need to know the relative address of the TLB datum within the chunk.

Is that it?

stskeeps:
> Is that it?

Yes

src/shadow-tlb.h
```diff
@@ -109,6 +109,19 @@ static inline uint64_t tlb_get_entry_cold_abs_addr(uint64_t eidx) {
         (ETYPE * sizeof(std::array<tlb_cold_entry, PMA_TLB_SIZE>)) + (eidx * sizeof(tlb_cold_entry));
 }
 
+template <TLB_entry_type ETYPE>
+static inline uint64_t tlb_get_entry_hot_rel_addr(uint64_t eidx) {
```
edubart, line 113: Can you simplify `tlb_get_entry_hot_abs_addr` to make use of this function, so this code is covered and not dead-code?

Then do the same for `tlb_get_entry_cold_abs_addr`.

edubart (changes requested):
I understood and I approve as long related code is rewritten to make use of them.

src/shadow-tlb.h
stskeeps: done! good idea

edubart (approved):

diegonehab (approved):

PR #89, opened 2023-06-29
fix: formatting in shadow-tlb.h

diegonehab:
We are still adjusting our CI. Somehow, this formatting error went through.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,5 @@
 name: Build/Test/Verify/Release
-on: [push, pull_request_target]
+on: [push, pull_request]
 env:
   ROM_VERSION: v0.16.0
   KERNEL_VERSION: v0.16.0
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -1,5 +1,5 @@
 name: Build/Test/Verify/Release
-on: [push, pull_request_target]
+on: [push, pull_request]
```
mpolitzer, line 2: Is this change security related?

vfusco: No, it was a misconfiguration. I misunderstood what the "context" meant. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

vfusco (approved):

mpolitzer (approved):

edubart (approved):

Issue #90, opened 2023-07-04
Build FDT directly in emulator and deprecate ROM


edubart:
## Context

Currently we maintain a ROM image where its sole goal is to build the FDT (flattened device tree). We could simplify and completely remove the need of a ROM image from the machine by building the FDT in emulator C++ code.

## Possible solutions

We can build the FDT using a FDT library in C++ directly on machine boot. We should take precautions to make the FDT deterministic if using an FDT library.

## Subtasks

- [x] Initialize entropy using FDT, make sure the initial entropy becomes a new machine config. This will allow to deprecate the need for `rndaddentropy` in tools.
- [x] Make sure FDT is deterministic.

PR #94, opened 2023-07-05
Build device tree directly into ROM

edubart:
This is the PR for https://github.com/cartesi/machine-emulator/issues/90


edubart:
This PR was included in https://github.com/cartesi/machine-emulator/pull/103

PR #95, opened 2023-07-07
Add proofs to uarch riscv tests

mpernambuco:
This branch adds the option to include proofs in the json log files produced by uarch-riscv-tests.lua.
The logged proofs will be used as test fixture by the Solidity step.

This PR also includes a fix to a bug that was found while testing the new option `--proofs-frequency=<number>`.
This bug occurs when proofs are not requesdted, leaving the merkle tree inconsistent because updated pmas are not marked dirty, causing the following exception:
```
merkle tree must be consistent when stepping alternating with and without proofs...
lua5.4: tests/machine-bind.lua:809: inconsistent merkle tree
stack traceback:
	[C]: in method 'step_uarch'
	tests/machine-bind.lua:809: in local 'f'
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -258,7 +258,7 @@ jobs:
       - name: Create uarch json logs to be used  to test the Solidity based microarchitecture interpreter
         run: |
           mkdir -p /tmp/uarch-riscv-tests-json-logs
-          /opt/cartesi/bin/uarch-riscv-tests --test-path=/opt/cartesi/tests --output-dir=/tmp/uarch-riscv-tests-json-logs json-logs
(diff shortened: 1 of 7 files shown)


diegonehab:
Have you checked how much faster it would be to keep the `update_merkle_tree_page()` and `update_after_write()` and instead make the implicit call to `update_merkle_tree()` optional in `get_proof()` and `get_root_hash()`? I assume you have tests that compute a bunch of step logs. Maybe they could be considerably sped up?

diegonehab (dismissed):

edubart (approved):

diegonehab (approved):

PR #109, opened 2023-07-20
Changed default init location

mpolitzer:
rootfs should not overwrite the system `init`.
We must use a custom location to make this work.

This requires a rootfs with https://github.com/cartesi/machine-emulator-tools/pull/12.

### DIFF
--- a/src/cartesi-machine.lua
+++ b/src/cartesi-machine.lua
@@ -402,7 +402,7 @@ local memory_range_replace = {}
 local ram_image_filename = images_path .. "linux.bin"
 local ram_length = 64 << 20
 local rom_image_filename = images_path .. "rom.bin"
-local rom_bootargs = "console=hvc0 rootfstype=ext2 root=/dev/mtdblock0 rw quiet swiotlb=noforce"
+local rom_bootargs = "console=hvc0 rootfstype=ext2 root=/dev/mtdblock0 rw quiet swiotlb=noforce init=/opt/cartesi/bin/init"
 local rollup
 local uarch
 local rollup_advance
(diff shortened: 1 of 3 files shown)


src/cartesi-machine.lua
```diff
@@ -402,7 +402,7 @@ local memory_range_replace = {}
 local ram_image_filename = images_path .. "linux.bin"
 local ram_length = 64 << 20
 local rom_image_filename = images_path .. "rom.bin"
-local rom_bootargs = "console=hvc0 rootfstype=ext2 root=/dev/mtdblock0 rw quiet swiotlb=noforce"
+local rom_bootargs = "console=hvc0 rootfstype=ext2 root=/dev/mtdblock0 rw quiet swiotlb=noforce init=/opt/cartesi/bin/init"
```
edubart, line 405: There is a test in `machine-test.lua` that will also need this change.

edubart (commented):
I like the idea of explicit passing the init script in bootargs. I think this is fine to go. Just need to make CI pass.

src/cartesi-machine.lua
mpolitzer: Nice catch, adding it to: `src/tests/machine-test.lua:512`

mpolitzer: I'll need an updated RootFS for this to pass CI, it is in the works.

edubart (approved):

mpolitzer:
> I like the idea of explicit passing the init script in bootargs. I think this is fine to go. Just need to make CI pass.

I'll need an updated RootFS ext2 image for this to pass CI, it is in the works.

vfusco:
This changes were included on #113

PR #115, opened 2023-08-11
fix: Use Solidity style

mpernambuco:
- Use solidity style `enum` names
- Remove `auto`

### DIFF
--- a/src/uarch-step.cpp
+++ b/src/uarch-step.cpp
@@ -14,6 +14,8 @@
 // with this program (see COPYING). If not, see <https://www.gnu.org/licenses/>.
 //
 
+// NOLINTBEGIN(google-readability-casting)
+
 #include <stdexcept>
 
 #include "riscv-constants.h"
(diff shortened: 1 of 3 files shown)


stephenctw (approved):

src/uarch-step.cpp
```diff
@@ -1089,34 +1091,34 @@ static inline void executeInsn(UarchState &a, uint32 insn, uint64 pc) {
...
+        return UArchStepStatus::CycleOverflow;
     }
     // do not advance if machine is halted
     if (readHaltFlag(a)) {
-        return uarch_step_status::uarch_halted;
+        return UArchStepStatus::UArchHalted;
     }
     // execute next instruction
     auto pc = readPc(a);
     auto insn = readUint32(a, pc);
```
edubart, line 1107: `auto` still being used in the above lines.

edubart (changes requested):

src/uarch-step.cpp
stephenctw: Good catch! @mpernambuco you probably missed the second patch I sent you, could you add it back?

stephenctw (approved):

edubart (approved):

vfusco (approved):

PR #131, opened 2023-08-24
feat: forward application exit status

mpolitzer:
Parse exit status as encoded by OpenSBI


mpolitzer:
No change needed here to support this feature

PR #132, opened 2023-08-25
Add WebAssembly support with Emscripten toolchain

edubart:
This PR contains a series of commits improving how we download and use our dependencies, so the cartesi machine becomes easier to build on any platform, then finally adds support to compile `libcartesi` to WebAsssembly.

In summary:
- Replaced crypto++ with XKCP to calculate Keccak-256 hashes
  - Also add support for AVX2 instructions, so machine root hash calculation performance doubles in systems that supports it.
- Replaced system's `nlohman-json` with its downloaded version in third-party directory.
- Reworked how mongoose is downloaded and compiled to allow disabling unneeded features
- Updated mongoose to 7.11
- Replaced `libb64` with own base64 implementation inspired in LuaSocket's implementation
- Removed remote-cartesi-machine proxy
- Removed the need to link any Boost library, we now only use header-only libraries from Boost.
- Replaced system's `boost` with boost headers 1.83.0 downloaded in third-party directory.
- Removed uneeded dependencies from `apt-get install`
- Cleanup some unused variables in Makefile
- Made many Makefile adjustments to allow cross compiling to WebAssembly. This includes `TARGET_OS` and `TARGET_ARCH` options.
- Introduced `make wasm-package` to create a package of `libcartesi` to use in WebAssembly, added instructions to readme.


Makefile
```diff
@@ -169,15 +174,21 @@ help:
 	@echo '  uarch-tests                - build and run microarchitecture rv64i instruction tests'
 	@echo '  uarch-tests-with-linux-env - build and run microarchitecture rv64i instruction tests using the linux-env docker image'
 
+checksum:
+	@cd $(DEPDIR) && sha1sum -c shasumfile
```
vfusco, line 178: sha1sum -> shasum

vfusco (changes requested):

Makefile
edubart: I intentionally changed this, so we don't have to add `perl` when packaging for Linux distributions (which is odd), while `sha1sum` comes with `coreutils` package in Linux distributions and is more widely available.

Any motivation to keep using `shasum` ?

edubart:
This PR was replaced by https://github.com/cartesi/machine-emulator/pull/156 .
Regarding XKCP, we decided to not rely on it, in favor of simplicity and to depend less on complex third party external libraries.

PR #136, opened 2023-09-14
feat: rollup rework

mpolitzer:
This PR simplifies the cartesi machine rollup by removing `input-metadata`, `voucher_hashes` and `notice_hashes`.
- The `input-metadata` contents will be integrated into the rollup `input`, details are yet to be determined.
- `vouchers`, `notices` and other types of messages will now be encoded in userspace. The cartesi-machine and the linux driver will treat them as binary blobs.
- `yield` will now pass the blob size in its data field. This way we won't need to inspect the encoded data.
- yield/HTIF won't care about the details of the messages anymore, so we will simplify the defines as well.

## test

- You need a compatible kernel, such as: https://github.com/cartesi/linux/pull/4
- Then run:
```
./remote-cartesi-machine 127.0.0.1:5001 &
cartesi-machine.lua --remote-address=127.0.0.1:5001 --checkin-address=127.0.0.1:5002 \
  --rollup-advance-state=input_index_begin:0,input_index_end:1 \
  --rollup --flash-drive=label:test,filename:../../kernel/artifacts/linux-selftest-5.15.63-ctsi-2-v0.0.0.ext2 \
  -- /mnt/test/drivers/cartesi/rollup_read
```


src/cartesi-machine.lua
```diff
@@ -161,71 +158,42 @@ where options are:
...
-        of input %%i of epoch %%e.
-
-        notice_hashes (default: "epoch-%%e-input-%%i-notice-hashes.bin")
-        the pattern that derives the name of the file written for the notice
-        hashes of input %%i of epoch %%e.
-
-        report (default: "epoch-%%e-input-%%i-report-%%o.bin")
-        the pattern that derives the name of the file written for report %%o
-        of input %%i of epoch %%e.
+        output_hash (default: "input-%%i-output-hash.bin")
```
diegonehab, line 188: This is the merkle root hash of the hashes of all outputs generated for that input? Maybe use the name "outputs-root-hash" instead of just "output-hash"?

src/cartesi-machine.lua
```diff
@@ -1504,59 +1425,41 @@ local function instantiate_filename(pattern, values)
...
-local function load_rollup_input_and_metadata(machine, config, advance)
+local function load_rollup_input(machine, config, advance)
     local values = { e = advance.epoch_index, i = advance.next_input_index }
-    machine:replace_memory_range(config.input_metadata) -- clear
-    load_memory_range(machine, config.input_metadata, instantiate_filename(advance.input_metadata, values))
     machine:replace_memory_range(config.rx_buffer) -- clear
-    load_memory_range(machine, config.rx_buffer, instantiate_filename(advance.input, values))
-    machine:replace_memory_range(config.voucher_hashes) -- clear
-    machine:replace_memory_range(config.notice_hashes) -- clear
+    return load_memory_range(machine, config.rx_buffer, instantiate_filename(advance.input, values))
```
diegonehab, line 1440: Ah, so much cleaner. :smile:

src/cartesi-machine.lua
```diff
@@ -1708,70 +1595,59 @@ else
...
                 machine:snapshot()
-                load_rollup_input_and_metadata(machine, config.rollup, rollup_advance)
+                local input_length = load_rollup_input(machine, config.rollup, rollup_advance)
                 if rollup_advance.hashes then print_root_hash(machine) end
                 machine:reset_iflags_Y()
-                machine:write_htif_fromhost_data(0) -- tell machine it is an rollup_advance state, but this is default
-                rollup_advance.voucher_index = 0
-                rollup_advance.notice_index = 0
-                rollup_advance.report_index = 0
+                machine:write_htif_fromhost_data((0 << 32) | input_length) -- tell machine it is an rollup_advance state, but this is default
```
diegonehab, line 1619: If we add the constant to `htif-defines.h` and to the Lua bind,  we could use `cartesi.machine.HTIF_YIELD_REASON_ADVANCE_STATE` here instead of `0`.

src/cartesi-machine.lua
```diff
@@ -1708,70 +1595,59 @@ else
...
-                end
                 -- there is an inspect state query to feed
                 if rollup_inspect and rollup_inspect.query then
                     stderr("\nBefore query\n")
-                    load_rollup_query(machine, config.rollup, rollup_inspect)
+                    local input_length = load_rollup_query(machine, config.rollup, rollup_inspect)
                     machine:reset_iflags_Y()
-                    machine:write_htif_fromhost_data(1) -- tell machine it is an inspect state
-                    rollup_inspect.report_index = 0
+                    machine:write_htif_fromhost_data((1 << 32) | input_length) -- tell machine it is an inspect state
```
diegonehab, line 1628: If we add the constant to `htif-defines.h` and to the Lua bind,  we could use `cartesi.machine.HTIF_YIELD_REASON_INSPECT_STATE` here instead of `1`.

diegonehab (changes requested):
I don't think we are removing exceptions and reports yet. So let's keep the names. Also, see comments in code.

src/cartesi-machine.lua
```diff
@@ -1656,9 +1548,6 @@ else
         assert(remote_address, "rollup requires --remote-address for snapshot/rollback")
         check_rollup_memory_range_config(config.rollup.tx_buffer, "tx-buffer")
         check_rollup_memory_range_config(config.rollup.rx_buffer, "rx-buffer")
-        check_rollup_memory_range_config(config.rollup.input_metadata, "input-metadata")
-        check_rollup_memory_range_config(config.rollup.voucher_hashes, "voucher-hashes")
```
edubart, line 1660: Do search all for `voucher_hashes` and `voucher-hashes`, there are still references for it in:
- `cartesi-machine.lua`
- `rollup-memory-range.lua`
- `htif-rollup.lua`
- `machine-bind.lua`

We could remove them right? Same for `notice_hashes` and `input_metadata`.

src/cartesi-machine.lua
```diff
@@ -1458,11 +1380,15 @@ local function check_rollup_htif_config(htif)
     assert(htif.yield_automatic, "yield automatic must be enabled for rollup")
 end
 
+local function set_yield_data(machine, reason, data)
+    machine:write_htif_fromhost_data(((reason & 0xffff) << 32) | (data & 0xffffff))
```
edubart, line 1384: I was confused why `set_yield_data` does `data & 0xffffff` while `get_yield` does `data & 0xffffffff`, did a `f` was missing here?

edubart (changes requested):

src/cartesi-machine.lua
```diff
@@ -176,69 +173,42 @@ where options are:
...
         input:<filename-pattern>
-        input_metadata:<filename-pattern>
         input_index_begin:<number>
         input_index_end:<number>
-        voucher:<filename-pattern>
-        voucher_hashes: <filename>
-        notice:<filename-pattern>
-        notice_hashes: <filename>
-        report:<filename-pattern>
+        output:<filename-pattern>
```
diegonehab, line 185: I just realized something. We are keeping notices/vouchers separate in the server manager for this release. I think we should also keep them separate here. I know this is sad, but I think it will be less confusing.

mpolitzer: So sad :cry:, separate: `report`, `notice` and `voucher` then.

diegonehab: Yeah. 😭

mpolitzer: This decision got reverted back to a single `output` of verifiable data, as per [output unification](https://discord.com/channels/600597137524391947/1172230296066871407) discussions.

tests/machine/src/htif_cmio.S
```diff
@@ -16,9 +16,9 @@
```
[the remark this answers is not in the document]
mpolitzer: thanks, removed.

src/cartesi-machine.lua
```diff
@@ -2007,83 +1969,63 @@ local function load_memory_range(machine, config, filename)
...
     local f = assert(io.open(name, "wb"))
-    -- skip offset to reach payload length
-    local length = string.unpack(">I8", machine:read_memory(config.start + 2 * 32 - 8, 8))
-    -- add offset and payload length to amount to be read
-    length = length + 2 * 32
     assert(f:write(machine:read_memory(config.start, length)))
     f:close()
 end
 
+local function check_outputs_root_hash(root_hash, output_hashes)
```
mpolitzer, line 2000: `make check-lua` returned:
```
cartesi-machine.lua:2001:51: shadowing upvalue output_hashes on line 30
```

src/cartesi-machine.lua
```diff
@@ -2007,83 +1969,63 @@ local function load_memory_range(machine, config, filename)
...
+    local z = string.rep("\0", 32)
+    if #output_hashes == 0 then
+        output_hashes = { z }
+    end
+    for i = 1, 64 do
```
mpolitzer, line 2005: Suppress `cartesi-machine.lua:2004:9: unused loop variable i` warning with:

```suggestion
    for _ = 1, 64 do
```

tools/template/tests-data-copyright.template
```diff
@@ -21,7 +21,6 @@ Files: /usr/share/cartesi-machine/tests/data/machine/access.*
        /usr/share/cartesi-machine/tests/data/machine/funary.*
        /usr/share/cartesi-machine/tests/data/machine/htif_console.*
        /usr/share/cartesi-machine/tests/data/machine/htif_invalid_ops.*
-       /usr/share/cartesi-machine/tests/data/machine/htif_rollup.*
```
mpolitzer, line 24: +`/usr/share/cartesi-machine/tests/data/machine/htif_cmio.*`?

src/cartesi-machine.lua
```diff
@@ -1999,60 +1936,62 @@ local function ilog2(value)
...
         is_power_of_two(range.length),
-        string.format("rollup range %s length not a power of two (%u)", name, range.length)
+        string.format("cmio range %s length not a power of two (%u)", name, range.length)
     )
     local log = ilog2(range.length)
     local aligned_start = (range.start >> log) << log
     assert(
         aligned_start == range.start,
-        string.format("rollup range %s start not aligned to its power of two size", name)
+        string.format("cmio range %s start not aligned to its power of two size", name)
```
mpernambuco, line 1950: Minor thing here.
This assert is not doing anything, I think.
The current implementation of ilog2() always returns 0.

mpernambuco (approved):
Minor preexisting issue in the ilog2() implementation. 
Won't be a problem after the load_cmio_input branch is merged because it will use fixed memory ranges.

src/cartesi-machine.lua
diegonehab:
```
local function ilog2(value)
    value = assert(math.tointeger(value), "expected integer")
    assert(value ~= 0, "expected non-zero integer")
    local log = 0
    while value > 1 do
        log = log + 1
        value = value >> 1
    end
    return log
end
```

diegonehab: I'll push a commit there with this fix.

diegonehab: Done.

tools/template/tests-data-copyright.template
edubart, line 1: This file has references to the old kernel `linux-6.5.9-ctsi-1-v0.19.1.bin` and old tools `rootfs-tools-v0.14.1`. Could we drop the use of versioned file names here? Like
`linux-6.5.9-ctsi-1-v0.19.1.bin` -> `linux.bin`.
`rootfs-tools-v0.14.1` -> `rootfs.bin`.

tests/Makefile
```diff
@@ -169,6 +176,15 @@ $(TARGETS_WITH_TOOLCHAIN):
 build-tests-misc-with-builder-image:
 	$(MAKE) emulator-builder-exec CONTAINER_COMMAND="make -j\$$(nproc) build-tests-misc"
 
+create-machines $(CARTESI_CMIO_PATH): | $(CARTESI_IMAGES)
```
edubart, line 179: If I type `make build-tests-all` twice, it will fail the second time here. I think doing a `rm -rf $(CARTESI_CMIO_PATH)` could fix this.

src/machine-config.h
```diff
@@ -167,29 +167,26 @@ using virtio_device_config = std::variant<virtio_console_config, ///< Console
...
-/// \brief Rollup configuration
-struct rollup_config {                                       // NOLINT(bugprone-exception-escape)
-    memory_range_config rx_buffer{0x60000000, 2 << 20};      ///< RX buffer
-    memory_range_config tx_buffer{0x60200000, 2 << 20};      ///< TX buffer
-    memory_range_config input_metadata{0x60400000, 4096};    ///< Buffer for input metadata
-    memory_range_config voucher_hashes{0x60600000, 2 << 20}; ///< Buffer for the voucher hash array
-    memory_range_config notice_hashes{0x60800000, 2 << 20};  ///< Buffer for the notice hash array
+/// \brief Cmio configuration
+struct cmio_config {                                    // NOLINT(bugprone-exception-escape)
+    memory_range_config rx_buffer{0x60000000, 2 << 20}; ///< RX buffer
```
edubart, line 172: We should use a constants here, instead of hard-coded value.
Same for the buffer size, so it will be easier to change this in the future.

Actually I see 0x60000000 and 0x60200000 in a some test Lua scripts, maybe we could expose a Lua variable with these value it.

tests/lua/cmio-test.lua
```diff
@@ -0,0 +1,367 @@
...
+        return assert(remote.machine(MACHINES_DIR .. name, runtime))
+    else
+        return assert(cartesi.machine(MACHINES_DIR .. name, runtime))
+    end
+end
+
+local function set_yield_data(machine, reason, data)
+    local m16 = (1 << 16) - 1
+    local m32 = (1 << 32) - 1
+    machine:write_htif_fromhost_data(((reason & m16) << 32) | (data & m32))
```
edubart, line 134: I am starting to think maybe we should have a specialized function to encode/decode yield data/reason. All this bit manipulation is a little fragile and being repeated in multiple places. I have noticed we did mistakes encoding/decoding this data field.

But not now, maybe in we could create an issue to do this later What you think @diegonehab ?
Could be something like `write_htif_yield_reason` same for `data`.

edubart (changes requested):
I reviewed and only found minor issues while testing locally.

tests/Makefile
```diff
@@ -56,6 +58,11 @@ ifeq ($(coverage),yes)
 NUM_JOBS = 1
 endif
 
+CREATE_MACHINES_OPTS ?=
+ifeq ($(rollup_init),yes)
+CREATE_MACHINES_OPTS += --rollup-init
```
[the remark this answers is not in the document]
diegonehab: I don't think this was ever an option of `cartesi-machine.lua`. Is it an option of some other program?

src/machine-config.h
diegonehab: Perna has a PR where these numbers are not configurable anymore and are exposed to Lua. I think it is still WIP and I am sure he has a more recent push in his computer. But this should be it. https://github.com/cartesi/machine-emulator/pull/192

tests/Makefile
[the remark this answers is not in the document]
mpolitzer: This was an option of a script imported from server manager.
I made this option obsolete by always creating both templates versions (ioctl-loop and echo-dapp) but forgot to remove it on the Makefile. Will do.

tests/Makefile
mpolitzer: Fixed `build-tests-all` dependency. It should build the templates only once now.

tests/lua/cmio-test.lua
edubart: Indeed, I forgot about it.

tools/template/tests-data-copyright.template
mpolitzer: Thanks, I'll update the PR.

My take on this is that versions on file names are not worth it, they increase the maintainability burden such as in this case and complicate CI. Also, Github already puts the assets below a tag on the download path. The benefit is allowing multiple versions to be downloaded without needing to rename them, which I think is not worth it.

Last time this was discussed, versions on filenames won, but we could discuss this further on discord or a meeting.

diegonehab: Was it Danilo that needed version names on files?

edubart (approved):

diegonehab (approved):

mpernambuco (approved):

PR #138, opened 2023-10-03
Remove ROM and write devicetree directly to new DTB memory range

diegonehab:
This PR cherry-picks the changes from a different PR by @edubart.
The central idea is the elimination of the ROM we used to build the device tree.
This is now built directly by the emulator, into a new memory range we call DTB.
The new memory range appears right before RAM and, at 1MiB, will enable more complex initialization of the Cartesi Machine.

Depends on https://github.com/cartesi/grpc-interfaces/pull/23  and https://github.com/cartesi/machine-emulator-defines/pull/8.


src/tests/util.lua
```diff
@@ -324,94 +217,48 @@ function test_util.parse_pma_file(filename)
...
-local PMA_PAGE_SIZE_LOG2 = 12
-local PMA_PAGE_SIZE = 1 << PMA_PAGE_SIZE_LOG2
+local PAGE_LOG2_SIZE = 12
+local PAGE_SIZE = 1 << PAGE_LOG2_SIZE
+local WORD_LOG2_SIZE = 3
 
-local function ceil_log2(x)
-    assert(x > 0, "shouldn't try to compute log of x <= 0")
-    return math.ceil(math.log(x, 2)) // 1
+local function merkle_hash(data, start, log2_size)
```
edubart, line 224: I think this function and also the `back_merkle_tree_meta` could be useful in the public Lua API, we could move it to a module in `cartesi` Lua library, the code will also become more isolated from these test utils.

src/tests/util.lua
```diff
@@ -324,94 +217,48 @@ function test_util.parse_pma_file(filename)
...
+    if log2_size > WORD_LOG2_SIZE then
```
edubart, line 225: Checking if we are trying to compute merkle hash of a page of zeros can speed up this function by 7x factor here, and this speed up our tests. Ex:

```lua
local ZERO_PAGE = string.rep('\x00', PAGE_SIZE)
local function merkle_hash(data, start, log2_size)
    if log2_size == PAGE_LOG2_SIZE and data:sub(start + 1, start + (1 << log2_size)) == ZERO_PAGE then
        return zero_keccak_hash_table[PAGE_LOG2_SIZE]
    end
```

edubart (dismissed):
I reviewed most files, code and tested, looks good for me, made two comments of possible enhancements (not mandatory though).

edubart (approved):

vfusco (changes requested):

edubart (approved):

Issue #93, opened 2023-07-04
Add a way to execute a large set of privileged commands on machine initialization


edubart:
## Context

Some command line options need to execute custom commands as root on boot, such as mounting a flash drive, initializing clock date and time, and initializing network. This feature is needed in the context of Hypervisor to boot hypervisor VMs, also in the context of VirtIO to properly initialize its drivers.

There is also a second problem about the limit of characters in the dapp command line that we must solve, because currently we cannot have a command line with more than ~4096 characters.

## Possible solutions

We can either use an extra flash drive for supplying the commands or build them directly in the FDT.

## Definitions of Done

- [x] Add new command line option to execute large sequence of commands (or shell script) as root user
- [x] Add new command line option to execute large sequence of commands (or shell script) as dapp user
- [x] Make sure command line option can work with large amount of characters
- [x] Simplify init script from tools, move some of its init to emulator, to make them optional
- [x] Add options to suppress some initialization features currently always done in init, such as mounting flash drives

Issue #70, opened 2023-05-31
Replace MTD flash drives with a better solution


edubart:
## Context

Currently we use MTD devices for flash drives to make their filesystems available inside the Linux machine. While it works, MTD has an overhead of its own and we could research a better and more lightweight solution. Ideally we should research something that could bypass the kernel page-cache mechanism, this is doable with any device that supports kernel [DAX](https://www.kernel.org/doc/Documentation/filesystems/dax.txt) (Direct Access for files).

For those unfamiliar to the kernel page-cache subject, every-time something is read/written to flash drive the kernel does not make the changes or read directly to the device, it usually reads/writes to a page, and eventually the kernel synchronizes this page at will. The kernel typically does this because device read/writes on actual hardware can be costly, however for a virtual machine that maps devices directly to host memory this is unnecessary and only downgrades the virtual machine performance.

## Possible solutions

One promising device that supports DAX are the [NVDIMM PMEM](https://access.redhat.com/documentation/pt-br/red_hat_enterprise_linux/7/html/storage_administration_guide/ch-persistent-memory-nvdimms) devices, sadly they do not support DAX on RISC-V yet, this mean we are unable to bypass yet page-cache. However PMEM still is simpler and faster than MTD, and we could completely replace MTD flash drives with it. Hopefully in the future RISC-V will start supporting DAX for it, so it's also more future proof.

There may be other options I am not aware of to bypass page-cache, this requires more research.

## Definition of Done

- [x] Replace entries of MTD with PMEM in FDT
- [x] Enable kernel configs related to NVIDMM PMEM
- [x] Adjust bootargs

PR #140, opened 2023-10-05
New init system

edubart:
This PR implements:
- #70
- #93 

This PR also fixes:
- https://github.com/cartesi/machine-emulator-tools/issues/6
- https://github.com/cartesi/machine-emulator-tools/issues/9
- https://github.com/cartesi/machine-emulator-tools/issues/10

This PR requires a new kernel, new init script and new rootfs.

## Depends on

- https://github.com/cartesi/grpc-interfaces/pull/18
- https://github.com/cartesi/image-kernel/pull/25
- https://github.com/cartesi/machine-emulator-tools/pull/16
- new rootfs (TODO)


src/cartesi-machine.lua
```diff
@@ -341,6 +354,25 @@ where options are:
     suppress cartesi-machine.lua output.
     exceptions: --initial-hash, --final-hash and text emitted from the target.
 
+  --no-init-splash
+    don't show cartesi machine splash on boot.
+
+  --append-init=<string>
+    append a command to machine's init script to be executed with root privilege,
```
diegonehab, line 361: End with a period, right?

src/cartesi-machine.lua
```diff
@@ -341,6 +354,25 @@ where options are:
...
+    the command is executed on boot after mounting flash drives and before running the entrypoint,
+    you can pass this option multiple times.
+
+  --append-init-script=<filename>
+    like --append-init, but use contents from a script file.
+
+  --append-entrypoint=<string>
+    append a command to machine's entrypoint script to be executed with dapp privilege,
```
diegonehab, line 369: End with a period, right?

src/cartesi-machine.lua
```diff
@@ -341,6 +354,25 @@ where options are:
...
 
+  --no-init-splash
+    don't show cartesi machine splash on boot.
+
+  --append-init=<string>
+    append a command to machine's init script to be executed with root privilege,
+    the command is executed on boot after mounting flash drives and before running the entrypoint,
+    you can pass this option multiple times.
+
+  --append-init-script=<filename>
```
diegonehab, line 365: Can we change "script" to "from-file"?

src/cartesi-machine.lua
```diff
@@ -119,6 +121,17 @@ where options are:
         target modifications to flash drive modify image file as well.
         by default, image files are not modified and changes are lost.
 
+        mount (optional)
+        whether the flash drive should be mounted automatically in init.
+        by default, the drive is mounted if there is an image file backing it,
+        you can use "mount:false" to disables auto mounting,
+        you can also use "mount:<path>" to choose a custom mount point.
+
+        user (optional)
```
diegonehab, line 130: Add comment on the default?

src/cartesi-machine.lua
```diff
@@ -1358,20 +1453,61 @@ else
...
             shared = flash_shared[label],
             start = flash_start[label],
             length = flash_length[label] or -1,
         }
+        -- auto mount
+        local mount = flash_mount[label]
+        local chownpath = "/dev/" .. devname
+        if label ~= "root" and mount then
+            local cmd = ([[busybox mkdir -p "MOUNT" && busybox mount /dev/DEVNAME "MOUNT"]])
+                :gsub("MOUNT", mount)
```
diegonehab, line 1488: Can the user pick MOUT=DEVNAME? :smile:

src/cartesi-machine.lua
```diff
@@ -401,8 +401,8 @@ local memory_range_replace = {}
 local ram_image_filename = images_path .. "linux.bin"
 local ram_length = 64 << 20
 local dtb_image_filename = nil
-local dtb_bootargs = "console=hvc0 rootfstype=ext2 root=/dev/mtdblock0 rw quiet \z
-                      swiotlb=noforce init=/opt/cartesi/bin/init random.trust_bootloader=on"
```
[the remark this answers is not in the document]
diegonehab: Ah, you added directly to the kernel, right?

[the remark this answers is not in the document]
edubart: Correct, I enabled this in the kernel config, so we can simplify the bootargs.

src/cartesi-machine.lua
edubart: Ok, but I changed to just `--append-init-file`, so without "from", because very long options are not user friendly.

src/cartesi-machine.lua
```diff
@@ -341,6 +354,25 @@ where options are:
...
+
+  --append-init-script=<filename>
+    like --append-init, but use contents from a script file.
+
+  --append-entrypoint=<string>
+    append a command to machine's entrypoint script to be executed with dapp privilege,
+    the command is executed after the machine is initialized and before the final entrypoint command,
+    you can pass this option multiple times.
+
+  --append-entrypoint-script=<filename>
```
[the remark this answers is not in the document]
edubart: Ok, but I changed to just `--append-init-file`, so without "from", because very long options are not user friendly.

src/cartesi-machine.lua
edubart: Good catch, I fixed the issue in case user picks that name :) , howeever I didn't went further with a concat across multiple lines.

src/cartesi-machine.lua
```diff
@@ -437,9 +437,9 @@ local memory_range_replace = {}
 local ram_image_filename = images_path .. "linux.bin"
 local ram_length = 64 << 20
 local dtb_image_filename = nil
-local dtb_bootargs = "console=hvc0 rootfstype=ext2 root=/dev/pmem0 rw quiet \z
-                      swiotlb=noforce init=/opt/cartesi/bin/init"
```
edubart: I checked that `swiotlb=noforce` is actually not needed anymore to workaround the memory usage by kernel itself when RAM >= 128MB, so I decided to remove it to simplify bootargs.

diegonehab (dismissed):

src/cartesi-machine.lua
diegonehab: So, this was only an issue when memory was small, right? We didn't want to waste memory reserving it for DMA stuff that we never use. When memory is greater than 128 they do not reserve? Is that the change? Or when memory is less than 128 they don't reserve?

src/cartesi-machine.lua
```diff
@@ -606,6 +649,26 @@ local options = {
...
+    {
+        "^%-it$",
+        function(all)
+            if not all then return false end
+            htif_console_getchar = true
+            local term, lang, lc_all = os.getenv("TERM"), os.getenv("LANG"), os.getenv("LC_ALL")
+            if term then append_init = append_init .. "export TERM=" .. term .. "\n" end
+            if lang then append_init = append_init .. "export LANG=" .. lang .. "\n" end
+            if lc_all then append_init = append_init .. "export LC_ALL=" .. lc_all .. "\n" end
+            local stty <close> = assert(io.popen("stty size"))
```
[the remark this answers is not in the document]
edubart: Indeed, `stty` is available in `coreutils` package on my system. In case it is not available the option will fail, and the user should either install it to use `-it` or fallback to old `-i` option.

src/cartesi-machine.lua
mpolitzer: Issue appeared when ram was 128MB or larger.
@edubart what kernel did you try removing it from?

commit: https://github.com/cartesi/machine-emulator/commit/525e152b32f954f9a679b272e51012e860e5e00b
ref: http://xillybus.com/tutorials/iommu-swiotlb-linux

edubart: @mpolitzer I tried kernel 6.1.x and 5.15.63, with different ram sizes, and I could not trigger the issue reported in https://github.com/cartesi-corp/machine-emulator/issues/439 . The kernel config today has many differences than the one at the time, I'm not sure why is not happening anymore, but since it is not happening, I decided to simplify.

vfusco (dismissed):

mpolitzer (approved):

diegonehab (approved):

PR #143, opened 2023-10-10
Rollup enable by default

### LINKED ISSUE #102: Have `rollup` enabled by default
## Context

Having rollup disabled is the exception. We should reflect that in the command line arguments to `cartesi-machine`.

## Possible solutions

Before: rollup **disabled** by default, `--rollup` enables it.
After: rollup **enabled** by default, `--no-rollup` disables it.

Something to consider is the deprecation process.
Tools or scripts that have `--rollup` would like a time period to update without completely being broken after an update.
Something to consider.

vfusco:
closes #102

### DIFF
--- a/src/cartesi-machine.lua
+++ b/src/cartesi-machine.lua
@@ -168,19 +168,17 @@ where options are:
     semantics are the same as for the --flash-drive option with the following
     difference: start and length are mandatory.
 
-  --rollup
-    defines appropriate values for rollup-rx-buffer, rollup-tx-buffer,
+  --no-rollup
+    do not define values for rollup-rx-buffer, rollup-tx-buffer,
     rollup-input-metadata, rollup-voucher-hashes, rollup-notice hashes,
     and htif yield for use with rollups.
-    equivalent to the following options:
+    default defined values are equivalent to the following options:
 
     --rollup-rx-buffer=start:0x60000000,length:2<<20
     --rollup-tx-buffer=start:0x60200000,length:2<<20
     --rollup-input-metadata=start:0x60400000,length:4096
     --rollup-voucher-hashes=start:0x60600000,length:2<<20
     --rollup-notice-hashes=start:0x60800000,length:2<<20
-    --htif-yield-manual
-    --htif-yield-automatic
 
   --rollup-advance-state=<key>:<value>[,<key>:<value>[,...]...]
     advances the state of the machine through a number of inputs in an epoch
@@ -306,11 +304,11 @@ where options are:
     This option will copy TERM, LANG, LC_ALL environment variables from the host to the guest,
     allowing the use of true colors and special characters when the host terminal supports.
 
-  --htif-yield-manual
-    honor yield requests with manual reset by target.
+  --no-htif-yield-manual
+    do not honor yield requests with manual reset by target.
 
-  --htif-yield-automatic
-    honor yield requests with automatic reset by target.
+  --no-htif-yield-automatic
+    do not honor yield requests with automatic reset by target.
 
   --store=<directory>
     store machine to <directory>, where "%%h" is substituted by the
@@ -448,7 +446,13 @@ local init_splash = true
 local append_bootargs = ""
 local append_init = ""
 local append_entrypoint = ""
-local rollup
+local rollup = {
+    rx_buffer = { start = 0x60000000, length = 2 << 20 },
+    tx_buffer = { start = 0x60200000, length = 2 << 20 },
+    input_metadata = { start = 0x60400000, length = 4096 },
+    voucher_hashes = { start = 0x60600000, length = 2 << 20 },
+    notice_hashes = { start = 0x60800000, length = 2 << 20 },
+}
 local uarch
 local rollup_advance
 local rollup_inspect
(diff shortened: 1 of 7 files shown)


src/cartesi-machine.lua
```diff
@@ -607,33 +611,28 @@ local options = {
         end,
     },
     {
-        "^%-%-htif%-yield%-manual$",
+        "^%-%-no%-htif%-yield%-manual$",
         function(all)
             if not all then return false end
-            htif_yield_manual = true
+            htif_yield_manual = false
```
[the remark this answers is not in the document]
vfusco: Sure it won't work. The user will get an error while trying to use it. But isn't this true for all other misconfigured parameters? So I'm not sure I agree, as yield can be used and was used in the past without rollups.

I was in doubt if I should disable the `htif_yield_manual` and `htif_yield_automatic` on `--no-rollup`. Now I think I shouldn't as there is no other way to turn it on if the users want to.

src/cartesi-machine.lua
```diff
@@ -607,33 +611,28 @@ local options = {
...
             return true
         end,
     },
     {
-        "^%-%-htif%-yield%-automatic$",
+        "^%-%-no%-htif%-yield%-automatic$",
         function(all)
             if not all then return false end
-            htif_yield_automatic = true
+            htif_yield_automatic = false
```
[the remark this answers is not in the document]
vfusco: see previous comment

src/cartesi-machine.lua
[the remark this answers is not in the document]
mpolitzer: > I was in doubt if I should disable the htif_yield_manual and htif_yield_automatic on --no-rollup. Now I think I shouldn't as there is no other way to turn it on if the users want to.

This is what will happen after the PR got updated:
`--no-rollup` -> rollup driver won't load due
`--no-yield-automatic` -> yield with `progress` and rollup device will be broken (can `finish` but can't `write`).
`--no-yield-manual` -> rollup device will be broken (can `write` but can't `finish`).

We could require both yields types to be available in the rollup driver initialization.
The yield flags are available in the FDT and so something limilar to what is in the yield driver.

There is also this idea to limit the yield driver to only allow `progress` calls.
Other calls can only be performed via the rollup driver.
Currently, with yield access, a process can mess up the rollup protocol by doing random yields.

https://github.com/cartesi/linux/pull/4/commits/283657c4ea8fa85dd48ee6610961ac4e714a3396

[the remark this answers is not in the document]
diegonehab: If the user disables either of the yields in the command-line, but does not disable rollups as well, we will end up with a machine config that has rollup but does not have yields. The emulator should throw and exception from within the constructor of `machine` explaining the problem. Right?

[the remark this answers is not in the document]
diegonehab: There is a weirdness where the user enables the yield before disabling rollup... That will be hard to catch. Can we fix the issue as well?

src/cartesi-machine.lua
```diff
@@ -448,7 +446,13 @@ local init_splash = true
 local append_bootargs = ""
 local append_init = ""
 local append_entrypoint = ""
-local rollup
+local rollup = {
```
edubart, line 449: This only solves for `cartesi-machine.lua`, what about the C API `cm_get_default_config` ? Shoudn't we also make this the default there? cc @diegonehab

src/machine-config.h
```diff
@@ -112,17 +112,17 @@ struct htif_config final {
...
 
 /// \brief Rollup configuration
-struct rollup_config {                    // NOLINT(bugprone-exception-escape)
-    memory_range_config rx_buffer{};      ///< RX buffer
-    memory_range_config tx_buffer{};      ///< TX buffer
-    memory_range_config input_metadata{}; ///< Buffer for input metadata
-    memory_range_config voucher_hashes{}; ///< Buffer for the voucher hash array
-    memory_range_config notice_hashes{};  ///< Buffer for the notice hash array
+struct rollup_config {                                                          // NOLINT(bugprone-exception-escape)
+    memory_range_config rx_buffer{.start = 0x60000000, .length = 2 << 20};      ///< RX buffer
```
edubart, line 121: It compiles, but I am getting these warnings:
```c++
machine-config.h:121:35: warning: C++ designated initializers only available with ‘-std=c++20’ or ‘-std=gnu++20’ [-Wc++20-extensions]
  121 |     memory_range_config rx_buffer{.start = 0x60000000, .length = 2 << 20};      ///< RX buffer
      |                                   ^
machine-config.h:121:56: warning: C++ designated initializers only available with ‘-std=c++20’ or ‘-std=gnu++20’ [-Wc++20-extensions]
  121 |     memory_range_config rx_buffer{.start = 0x60000000, .length = 2 << 20};      ///< RX buffer
      |                                                        ^
```

Using `memory_range_config rx_buffer{0x60000000, 2 << 20};` should fix.

edubart (changes requested):

src/machine-config.h
diegonehab: I used it in LambaDEX, but I selected c++20. Victor must have seen it. Shall we make the switch? This is really cool. :)

src/cartesi-machine.lua
diegonehab: Yes. It should be default in C++ as well.

diegonehab: Victor will create an issue about 'get_default_config' vs. 'cartesi-machine.lua'

diegonehab (approved):

src/cartesi-machine.lua
vfusco: https://github.com/cartesi/machine-emulator/issues/153

Issue #96, opened 2023-07-07
Expose pma address ranges in Lua binding


mpernambuco:
## Context
The proliferation of hard-coded start address and length of pma ranges in Lua programs is an inconvenience that can be solved by exposing all relevant values in the Lua binding of the Cartesi Machine.

## Possible solutions
Expose in the Lua binding of the Cartesi Machine all  start addresses and lengths of pmas currently defined in pma-defines.h

diegonehab:
I think this issue is related to #47. The way I see it, we should have code that works like this:

```
for pma in m:get_pmas() do
    print(pma.start)
    print(pma.length)
    print(pma.name)
    print(pma.flags.DID)
    print(pma.flags.W)
    print(pma.flags.R)
    ...
end
```

and similar code for C++, C, and jsonrpc. What do you think?  That way you could do, in C++, for example

```
    const auto &pmas = m.get_pmas();
    if (const auto tlb = std::find_if(std:begin(pmas), std::end(pmas), [](auto &pma) { return pma.DID == DID_TLB; }); it != std::end(pmas)) {
        std::cerr << tlb->start << std::endl;
    }
```

Or some such.

PR #148, opened 2023-10-25
Memory range introspection

### LINKED ISSUE #47: Alternative ways to compute a machine's state hash
## Context

We need to make sure the machine Merkle tree implementation works correctly and indeed captures the entire state of the machine. It is therefore important that we be able to reproduce the root hash without using the code in the machine itself. 

## Possible solutions

At least two ways of doing this would be useful.

We should be able to, using the machine interface, iterate over the contents of the machine and externally compute the state hash. This would be useful in tests.

We should be able to, without instantiating a machine, and iterating over the contents of a machine config, compute the state hash. This would be useful to check integrity of a stored machine directory.

## Subtasks

- [ ] Ensure the machine API exposes enough information to allow the external computation of the state hash
- [ ] Implement a function that computes the state hash given a machine instance
- [ ] Implement a function  that computes the state hash given a machine config

diegonehab:
This closes #47 and #96. It depends on grpc-interfaces PR [#24](https://github.com/cartesi/grpc-interfaces/pull/24)

### DIFF
--- a/src/tests/machine-bind.lua
+++ b/src/tests/machine-bind.lua
@@ -492,7 +492,7 @@ do_test("should return expected value", function(machine)
     print("Root hash: ", test_util.tohex(root_hash))
 
     machine:dump_pmas()
-    local calculated_root_hash = test_util.calculate_emulator_hash(pmas_file_names)
+    local calculated_root_hash = test_util.calculate_emulator_hash(machine)
     for _, file_name in pairs(pmas_file_names) do
         os.remove(test_path .. file_name)
     end
--- a/src/tests/machine-bind.lua
+++ b/src/tests/machine-bind.lua
@@ -267,18 +267,6 @@ local function connect()
     return remote, version
 end
 
-local pmas_file_names = {
-    "0000000000000000--0000000000001000.bin", -- shadow state
-    "0000000000010000--0000000000001000.bin", -- shadow pmas
-    "0000000000020000--0000000000006000.bin", -- shadow tlb
-    "0000000002000000--00000000000c0000.bin", -- clint
-    "0000000040008000--0000000000001000.bin", -- htif
-    "000000007ff00000--0000000000100000.bin", -- dtb
-    "0000000080000000--0000000000100000.bin", -- ram
-    "0000000070000000--0000000000010000.bin", -- uarch ram
-}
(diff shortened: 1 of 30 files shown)


src/tests/util.lua
```diff
@@ -219,10 +219,8 @@ end
 function test_util.align(v, el) return (v >> el << el) end
 
 function test_util.load_file(filename)
-    local fd = assert(io.open(filename, "rb"))
-    local data_size = fd:seek("end")
-    fd:seek("set")
-    local data = fd:read(data_size)
+    local fd <close> = assert(io.open(filename, "rb"))
+    local data = assert(fd:read("*all"))
     fd:close(filename)
```
edubart, line 224: Unnecessary close

src/tests/machine-test.lua
```diff
@@ -277,63 +237,44 @@ test_util.make_do_test(build_uarch_machine, machine_type)(
...
-
-    local ram_address_start = tonumber(test_util.split_string(ram_file_name, "--.")[1], 16)
-    local ram_log2_size = math.ceil(math.log(#ram, 2))
-    local calculated_ram_hash = test_util.merkle_hash(ram, 0, ram_log2_size)
-
-    local ram_proof = machine:get_proof(ram_address_start, ram_log2_size)
+    -- find ram memory range
+    local ram
+    for _, v in ipairs(machine:get_memory_ranges()) do
+        print(v.description)
```
edubart, line 263: Unnecessary print

src/tests/machine-bind.lua
```diff
@@ -545,7 +541,7 @@ end)
 print("\n\n dump pmas to files")
 do_test("there should exist dumped files of expected size", function(machine)
     -- Dump pmas to files
-    machine:dump_pmas()
+    machine:dump_memory_ranges()
 
     for i = 1, #pmas_file_names do
```
edubart, line 546: This hardcoded list of pmas could be moved inside this test. Morever the `pmas_sizes` could be extracted from `pmas_file_names`. Or the pmas could be listed from `get_memory_ranges`.

diegonehab:
I'll just get rid of the method everywhere.

edubart (approved):

edubart (dismissed):

edubart (dismissed):

edubart (approved):

vfusco (approved):

edubart (approved):

PR #150, opened 2023-10-27
refactor: replace libb64 with custom base64 implementation

edubart:
This is to simplify our dependencies, so the cartesi machine is easier to build on more platforms.

The base64 implementation was copied from LuaSocket code which @diegonehab himself wrote.


src/base64.cpp
```diff
@@ -14,36 +14,152 @@
...
+        ctxlen = b64encode(static_cast<uint8_t>(b), ctx, ctxlen, sout);
+        if (ctxlen == 0) { // appended 4 characters
+            ncols += 4;
+            if (ncols >= 72) { // add CRLF every 72 columns
+                sout << "\r\n";
+                ncols = 0;
+            }
+        }
+    }
+    b64pad(ctx, ctxlen, sout);
```
diegonehab, line 152: Is there a way for the padding to blow up the 72 col limit? I think so, right? So perhaps add another if before padding?

src/base64.cpp
```diff
@@ -14,36 +14,152 @@
...
+    }
+    return 0;
+}
+
+// Acumulates bytes in input buffer until 4 bytes are available.
+// Translate the 4 bytes from Base64 form and append to buffer.
+// Returns new number of bytes in buffer.
+static size_t b64decode(uint8_t c, uint8_t *input, size_t size, std::ostringstream &sout) {
+    // ignore invalid characters
+    if (b64unbase[c] > 64) {
```
diegonehab, line 105: I think we should skip whitespaces but  throw an error otherwise.

diegonehab (changes requested):
See comments.

src/base64.cpp
edubart: I don't think this is possible (I tried), I don't want to add a check for this if it's a branch never taken.

src/base64.cpp
edubart: Ok, I explicitly defined ` \t\n\v\f\r` as the valid set for whitespaces.

src/base64.cpp
diegonehab: It was a good question, but I agree it is not possible. The padding always rounds up the number of bytes in the output to a multiple of 4. Since 72 is also a multiple of 4, the padding can never be the reason a line goes over the limit.

src/base64.cpp
diegonehab: Since you went further, what do you think we should be implementing? The [MIME]( https://datatracker.ietf.org/doc/html/rfc2045) spec or the later [RFC](https://datatracker.ietf.org/doc/html/rfc4648). LuaSocket was used with MIME. But here we are not using MIME. In a nutshell, MIME requires encoding to split at most into 76 columns and asks decoding to ignore invalid characters. The later RFC says not to split into lines and to reject any invalid input. I guess we could stop breaking into lines when encoding but accept lines when decoding.

edubart: I would say to not break lines when encoding, to save bytes and to make easier for others to decode base64 generated from us. If you agree, I will make the change.

diegonehab (approved):

vfusco (approved):

PR #152, opened 2023-10-27
refactor: update mongoose and remove the need of boost filesystem

edubart:
This is to simplify our dependencies, so the cartesi machine is easier to build on more platforms.

Notes:
- I did some cleanups and refactoring on how mongoose is bundled.
- I refactored some code to not depend on linking any boost library, now the code base uses only headers-only libraries.
- I also removed some unneeded packages from Dockerfile and README.


diegonehab:
Boost and crypto are kind of unpleasant to bundle. The Google way is to use some crazy alternative to makefile, and give users the option to use the bundled or the system version of each dependency. I think this is overkill for us. Perhaps we could create a new repo for this self-contained effort? I really don't want to support building third-party software in every kind of possible system where they might be needed. I also don't want to burden CI and people that do not need this. :/ @vfusco ?

edubart:
> I really don't want to support building third-party software in every kind of possible system where they might be needed

It's important to note that Boost is not being built here, this PR makes the code base use header-only libraries from Boost. The only burden here is downloading and extracting boost, which is usually done once with `make dep`, and this PR only extract the boost headers not the whole tarball.

edubart:
I reworked the PR to not download Boost header, it will use the system boost headers.

diegonehab (dismissed):

vfusco (approved):

vfusco (approved):

vfusco (approved):

diegonehab (approved):

PR #156, opened 2023-11-07
feat: support cross compiling libcartesi as a static library

edubart:
Included changes:

- introduce the `TARGET_OS` make parameter, to allow cross compiling to other OSes
- make `termios.h` non-mandatory to compile libcartesi.a when defining `NO_TERMIOS`, because WASI SDK target has no support for it
- build `libcartesi.a` static library
- remove the use of `std::filesystem` from `libcartesi.a`, because WASI SDK target has no support for it
- refactor object file lists in the makefile, to include only what it should in the static library
- build and install `libcartesi_jsonrpc.so` and `libcartesi_jsonrpc.a`, the code was there, but it was not being compiled or installed
- removed all uses of `rpath` and `patchelf`, now all executables and shared libraries are statically linked
- `libcartesi.so` still is being built, but is not being linked by anyone, people will be able to choose to link to `libcartesi.so` or `libcartesi.a`
- refactored `make install` to install and strip properly the binaries and libraries
- some makefile cleanups
- added possibility to compile libcartesi.a without `mmap`
- isolated all OS related functions in `os.cpp` file with automatic detection of supported functions when targeting WASI or Windows
- added ways to disable all OS function implementations with `-DNO_TTY` for example, in case the target doesn't have an OS supporting that API
- the flags `-std=c++17 -pedantic` was replaced with `-std=gnu++17 -Wpedantic`. Because the old flags actually disabled some POSIX functions (such as `mkdir` and `gettimeofday`) in some toolchains (such as Cosmopolitan or DJGPP for DOSBox).
- the `privileged` and `PAGE_SIZE` keywords is a define in some toolchains and gave compile errors, so I renamed them
- added `make bundle-boost`, for making easy cross compiling libcartesi with other toolchains
- allow compiling `libcartesi.a` in standalone without need to have GRPC or Lua installed in the system, added instructions to README, this important when only wanting to cross compile with other toolchains
- refactored Makefile to remove some `:=`, to invoke some shell commands which depends on system having protobuf, grpc, lua etc only when really needed by a job
- removed `CM_API ` from `machine-c-api-internal.h` functions because those APIs should not be exposed in the public C API
- added support for compiling libcartesi in systems without threading support with `make nothreads=yes`
- fixed issues for building on MacOS with homebrew with old clang toolchain

After this PR, it is possible to compile a `libcartesi.a` for WASM using two different toolchains (as long you do `make bundle-boost` before):

With WASI SDK toolchain:
```sh
make -C src -j8 release=yes \
  TARGET_OS=Linux \
  CC=/opt/wasi-sdk/bin/clang \
  CXX=/opt/wasi-sdk/bin/clang++ \
  AR="/opt/wasi-sdk/bin/llvm-ar rcs" \
  libcartesi.a
```

With Emscripten toolchain:
```sh
make -C src -j8 release=yes \
  TARGET_OS=Linux \
  CC=emcc \
  CXX=em++ \
  AR="emar rcs" \
  libcartesi.a
```

With GCC RISC-V toolchain (for running a cartesi machine inside itself, yay):
```sh
make -C src -j8 release=yes \
  TARGET_OS=Linux \
  CC=riscv64-linux-gnu-gcc \
  CXX=riscv64-linux-gnu-g++ \
  AR="riscv64-linux-gnu-ar rcs" \
  libcartesi.a
```

With mingw-w64 toolchain (for Windows build):
```sh
make -C src -j8 release=yes \
    TARGET_OS=Linux \
    CC=x86_64-w64-mingw32-gcc \
    CXX=x86_64-w64-mingw32-g++ \
    AR="x86_64-w64-mingw32-ar rcs" \
    MYSOLDFLAGS="-llua.dll -static" \
    libcartesi.a cartesi.so
```

NOTE: `TARGET_OS=Linux` and `cartesi.so` seem suspicious, but is the correct way to cross compile to Windows build for now.

I also tested libcartesi of this PR to compile with the following platforms/toolchains:
- FreeBSD with its GCC toolchain
- DOS with DGJPP toolchain
- Windows with Visual Studio clang toolchain
- Comopolitan libc with its cosmocc toolchain (after removing exceptions)
- MacOS 10.15 x86_64 with clang 12 + Homebrew
- Android


src/Makefile
```diff
@@ -676,23 +645,25 @@ $(GROUND_TRUTH_DIR)/$(LOG_DIR)/%.json.br: $(CARTESI_TESTS_PATH)/%.bin $(GROUND_T
...
 
 remote-cartesi-machine: $(REMOTE_CARTESI_MACHINE_OBJS)
-	$(CXX) $(LDFLAGS) $(CARTESI_EXECUTABLE_LDFLAGS) -o $@ $(REMOTE_CARTESI_MACHINE_OBJS) $(REMOTE_CARTESI_MACHINE_LIBS)
+	$(CXX) -o $@ $(REMOTE_CARTESI_MACHINE_OBJS) $(REMOTE_CARTESI_MACHINE_LIBS) $(LDFLAGS) $(EXELDFLAGS)
 
 jsonrpc-remote-cartesi-machine: $(JSONRPC_REMOTE_CARTESI_MACHINE_OBJS)
-	$(CXX) $(LDFLAGS) $(CARTESI_EXECUTABLE_LDFLAGS) -o $@ $(JSONRPC_REMOTE_CARTESI_MACHINE_OBJS) $(JSONRPC_REMOTE_CARTESI_MACHINE_LIBS)
+	$(CXX) -o $@ $(JSONRPC_REMOTE_CARTESI_MACHINE_OBJS) $(JSONRPC_REMOTE_CARTESI_MACHINE_LIBS) $(LDFLAGS) $(EXELDFLAGS)
+
+$(PROTOBUF_GEN_OBJS) $(GRPC_GEN_OBJS) remote-cartesi-machine: CXXFLAGS += $(GRPC_PROTOBUF_INC)
```
vfusco, line 661: The build with macports broke. Some files are building without proper includes. The following change fixes for me:

```
$(LIBCARTESI_GRPC_OBJS) remote-cartesi-machine: CXXFLAGS += $(GRPC_PROTOBUF_INC)
```

vfusco (changes requested):

vfusco (approved):

src/Makefile
```diff
@@ -328,19 +326,22 @@ LIBCARTESI_OBJS:= \
...
 LUACARTESI_OBJS:= \
-	sha3.o \
 	clua-cartesi.o \
-	clua-i-virtual-machine.o \
 	clua-machine.o \
-	clua-htif.o \
-	clua-machine-util.o \
-	clua.o
+	$(CARTESI_CLUA_OBJS) \
+	$(LIBCARTESI_OBJS)
```
diegonehab, line 340: Shouldn´t LUACARTESI simply use LIBCARTESI?

src/Makefile
```diff
@@ -352,41 +353,39 @@ GRPC_GEN_OBJS:= \
...
-	pristine-merkle-tree.o \
-	$(GRPC_GEN_OBJS) \
-	$(PROTOBUF_GEN_OBJS) \
 	grpc-virtual-machine.o \
+	grpc-machine-c-api.o \
 	protobuf-util.o \
-	grpc-machine-c-api.o
+	$(GRPC_GEN_OBJS) \
+	$(PROTOBUF_GEN_OBJS) \
+	$(LIBCARTESI_OBJS)
```
diegonehab, line 357: Shouldn´t LUACARTESI_GRPC simply use LIBCARTESI?

src/Makefile
```diff
@@ -352,41 +353,39 @@ GRPC_GEN_OBJS:= \
...
+
+LIBCARTESI_JSONRPC_OBJS:= \
+	jsonrpc-virtual-machine.o \
+	jsonrpc-machine-c-api.o \
+	mongoose.o \
+	$(LIBCARTESI_OBJS)
```
diegonehab, line 363: Shouldn´t LUACARTESI_JSONRPC simply use LIBCARTESI?

src/Makefile
```diff
@@ -352,41 +353,39 @@ GRPC_GEN_OBJS:= \
...
 
 LUACARTESI_GRPC_OBJS:= \
-	clua-i-virtual-machine.o \
-	clua-htif.o \
-	clua-machine-util.o \
 	clua-cartesi-grpc.o \
 	clua-grpc-machine.o \
-	clua.o
+	$(CARTESI_CLUA_OBJS) \
+	$(LIBCARTESI_GRPC_OBJS)
```
diegonehab, line 369: Shouldn´t LUACARTESI_GRPC simply use LIBCARTESI_GRPC library, instead of the objects?

src/Makefile
```diff
@@ -352,41 +353,39 @@ GRPC_GEN_OBJS:= \
...
 	clua-cartesi-jsonrpc.o \
 	clua-jsonrpc-machine.o \
-	json-util.o \
-	base64.o \
-	mongoose.o \
-	jsonrpc-machine-c-api.o \
-	jsonrpc-virtual-machine.o \
-	clua.o
+	$(CARTESI_CLUA_OBJS) \
+	$(LIBCARTESI_JSONRPC_OBJS)
```
diegonehab, line 375: Shouldn´t LUACARTESI_JSONRPC simply use LIBCARTESI_JSONRPC library, instead of the objects?

src/Makefile
```diff
@@ -544,9 +549,8 @@ TEST_MERKLE_TREE_HASH_OBJS:= \
 
 TEST_MACHINE_C_API_OBJS:= \
     test-machine-c-api.o \
-	sha3.o \
     back-merkle-tree.o \
-    pristine-merkle-tree.o
+    $(LIBCARTESI_GRPC_OBJS)
```
diegonehab, line 561: Also use library

src/Makefile
```diff
@@ -557,67 +561,19 @@ $(PROTO_OBJS): CXXFLAGS +=  -Wno-zero-length-array -Wno-unused-parameter -Wno-de
...
-	json-util.o \
-	base64.o \
-	interpret.o \
-	uarch-machine.o \
-	uarch-step.o \
-	uarch-interpret.o
+	slog.o \
+	$(GRPC_GEN_OBJS) \
+	$(PROTOBUF_GEN_OBJS) \
+	$(LIBCARTESI_OBJS)
```
diegonehab, line 577: Also use library

src/Makefile
```diff
@@ -557,67 +561,19 @@ $(PROTO_OBJS): CXXFLAGS +=  -Wno-zero-length-array -Wno-unused-parameter -Wno-de
...
-	pristine-merkle-tree.o \
-	pma.o \
-	machine.o \
-	machine-config.o \
-	interpret.o \
-	uarch-machine.o \
-	uarch-step.o \
-	uarch-interpret.o
+	slog.o \
+	$(LIBCARTESI_OBJS)
```
diegonehab, line 584: Also use library.

Makefile
```diff
@@ -81,12 +83,12 @@ UARCH_TO_SHARE= uarch-ram.bin
 MONGOOSE_VERSION=7.12
 
 # Build settings
-DEPDIR := third-party
-SRCDIR := $(abspath src)
-DOWNLOADDIR := $(DEPDIR)/downloads
-DEPDIRS := third-party/mongoose-$(MONGOOSE_VERSION)
-SUBCLEAN := $(addsuffix .clean,$(SRCDIR) uarch third-party/riscv-arch-tests)
-COREPROTO := lib/grpc-interfaces/core.proto
+DEPDIR = third-party
```
diegonehab, line 86: Why change to "="? Does the value of any variable get reassigned between evaluations?

src/Makefile
edubart: It's almost the same thing, `libcartesi.a` is just a "zip" of all `.o` objects. By using objects directly instead of `.a` the build can skip the `ar rcs` step, so it's more error prone in case the user forget to ser `AR` variable when cross compiling and he wants just to build `.so` files.

src/Makefile
```diff
@@ -248,7 +248,7 @@ EMPTY:=
 SPACE:=$(EMPTY) $(EMPTY)
 CLANG_TIDY_HEADER_FILTER=$(PWD)/($(subst $(SPACE),|,$(LINTER_HEADERS)))
 
-CXXFLAGS+=$(OPTFLAGS) -std=c++17 -fvisibility=hidden -fPIC -MMD $(CC_MARCH) $(INCS) $(GCFLAGS) $(UBFLAGS) $(DEFS) $(WARNS)
+CXXFLAGS+=$(OPTFLAGS) -std=gnu++17 -fvisibility=hidden -fPIC -MMD $(CC_MARCH) $(INCS) $(GCFLAGS) $(UBFLAGS) $(DEFS) $(WARNS)
```
diegonehab: What crazyness required this?

Makefile
edubart: The use `:=` forces evaluation of commands once and immediately, while `=` only when the variable is instantiated. This is a problem when doing cross compilation and some command that is not available in the system could be skipped.

A good example is this line:
```
GRPC_PROTOBUF_LIB=$(shell pkg-config --libs grpc++ protobuf)
````
In case we used `:=` it would require invoking `pkg-config --libs grpc++ protobuf` even when not compiling grpc stuff, and throw compile errors.

I made a refactor to replace almost all `:=` with `=` to make the Makefile more cross compilation friendly.

src/Makefile
edubart: Some systems that I cross compiled (such as DOS), did not have all POSIX functions available. Using just `-std=c++17` did not enable non standard C OS APIs, such as `gettimeofday`, `mkdir`, and others, because they are not in standard C. I think some systems are free to disable POSIX APIs with `-std=c++17`.

So having `-std=gnu++17` is more friendly for cross compilation, another option could be removing `-std=` and let use the default one.

src/Makefile
edubart: I am reworking to *make use* of the static libraries.

src/Makefile
diegonehab: In an ideal world, we would change this flag only on os.cpp, right?

diegonehab (approved):

vfusco (approved):

Issue #134, opened 2023-09-08
Custom machine instructions


edubart:
## Context

Currently the only way to suspend the machine from inside to execute external in the outside is through the yield device with automatic yields, while this works it has some limitations. Triggering external code is useful for frontends that is embedding a cartesi machine emulator and wants to to dispatch commands (such as audio/video commands), dapps may want to embed a cartesi machine for deterministic computations in the frontend, so its computations always match the rollups backend.

Limitations of using automatic yields in the frontend:
- The round trip of a yield automatic consumes thousands cycles, and triggers RISC-V privilege changes and TLB shootdowns, if we think of a dapp that wants to dispatch audio/video commands in real-time in the frontend, this is costly. I am working around this by creating command queues and dispatching a few times per second, but this creates the burden of creating more complex code to handle command queues and command buffers between the outside "device" and inside "driver".
- The yield operation switches the RISC-V privilege from U-mode to M-mode, so when we are executing outside, we cannot read virtual memory from the running process anymore, because the process context is lost. This must be worked around by creating physically fixed mapped memory so the outside code can read, and to make this we have to create hacks to set the virtual memory to a fixed physical memory region, or even new linux devices just to expose virtual memory from a process externally. I am working around this by enabling HUGETLBFS in the kernel and allocating a 2MB page on boot, this page will be fixed in a physical memory region forever and I can map a 2MB virtual memory from an process to it, but this is hacky and I am limited to 2MB.
- The yield device requires privilege, sandboxed applications running untrusted code that want to be verified in a rollups backend should not have access to the yield device, it's a security risk, a dispatch operation can be executed as a no-op in the rollups backend, so using custom instructions is safer. I am working around this by not using the yield device in the rollups backend, because the instructions are no-op anyway, so the yield device is just used in the frontend, but this makes the code executed in the frontend/backend slightly different and hurts determinism.

A typical custom instruction I am talking about, would be something like draw a framebuffer, play an audio from a buffer, clear the screen, sleep for a few milliseconds, etc, notice that these custom instructions has only meaning in the frontend, while the backend can just ignore and execute as no-op.

So I want to propose adding a new way to add custom RISC-V instructions (or system calls), that trigger external code when a callback C handler is registered in the frontend, but do nothing when in rollups backend, because no handler will ever be registered.

[...]

PR #161, opened 2023-11-20
Add soft yield instruction

edubart:
This PR introduces the soft yield runtime config, it's a new kind of yield with the purpose of stopping the interpreter main loop immediately, to execute external code outside the machine.

This feature is only useful when using libcartesi library in applications that uses cartesi machine in its frontend, and has no use in the cartesi-machine cli or in the rollups backend, and it's disabled by default.

A soft yield happens when soft yield runtime config is true and the instruction `SRAIW x0, x31, shamt` is executed for any given `shamt` (32 possible values), the main interpreter loop will break with `CM_BREAK_REASON_YIELDED_SOFTLY` reason, and the programmer of the cartesi machine can handle this situation to perform operations in the outside.

When this instruction is executed in the rollups backend, it works as a NOP operation, while being a valid RISC-V instruction. SRAIW was chosen from the list of HINT instructions reserved from custom use in the RISC-V specification. Only shift instructions have HINTs available reserved for custom use, I choose SRAIW because I think it's the most uncommon one among the available shift instructions, to impact less the additional branching.

The main advantage of the soft yield when comparing with automatic yields are:
- It's faster, the interpreter breaks immediately when this instruction is reached, while the automatic yield takes a few thousands cycles
- The yield can happen in U-mode, this allows to read user space virtual memory from outside with `read_virtual_memory`, so there is no need to reserve fixed physical memory for reading user space memory from outside
- It performs as a lightweight no-op in the rollup backend, this can be useful for dapps that want to have a frontend emulator spitting out data (such as graphics, audio, text) in the frontend, while in the backend this is silently ignored because there is no frontend attached.
- It's unprivileged, there is no need to switch to M-mode or special devices to perform the soft yield, meaning no TLB flushes will happen
- It's a way to implement custom instructions in the frontend for users of the libcartesi

This is a solution for #134 , more disadvantages of automatic yields were also described there

## Depends on

- https://github.com/cartesi/grpc-interfaces/pull/28


src/grpc-virtual-machine.cpp
```diff
@@ -348,8 +348,9 @@ interpreter_break_reason grpc_virtual_machine::do_run(uint64_t mcycle_end) {
         return interpreter_break_reason::yielded_manually;
     } else if (response.iflags_x()) {
         return interpreter_break_reason::yielded_automatically;
+    } else if (response.mcycle() != mcycle_end) {
```
diegonehab, line 351: Hmmm. Should we give preference to the soft yield over mcycle_end? This way, the program controlling the emulator in a loop would receive the `yielded_softly`, deal with it, restart the emulator, and only then get the `reached_target_mcycle`. Otherwise, the soft yield is lost.

diegonehab: There is no flag for this soft yield. So, unlike the other yields, there is no other way for the program controlling the emulator to know.

src/interpret.cpp
```diff
@@ -5640,8 +5645,9 @@ interpreter_break_reason interpret(STATE_ACCESS &a, uint64_t mcycle_end) {
         return interpreter_break_reason::yielded_manually;
     } else if (a.read_iflags_X()) {
         return interpreter_break_reason::yielded_automatically;
-    } else {                                   // Reached mcycle_end
-        assert(a.read_mcycle() == mcycle_end); // LCOV_EXCL_LINE
+    } else if (a.read_mcycle() != mcycle_end) {
```
diegonehab, line 5648: Same comment here.

src/json-util.cpp
```diff
@@ -436,6 +436,7 @@ void ju_get_opt_field(const nlohmann::json &j, const K &key, machine_runtime_con
     ju_get_field(j[key], "htif"s, value.htif, path + to_string(key) + "/");
     ju_get_opt_field(j[key], "skip_root_hash_check"s, value.skip_root_hash_check, path + to_string(key) + "/");
     ju_get_opt_field(j[key], "skip_version_check"s, value.skip_version_check, path + to_string(key) + "/");
+    ju_get_opt_field(j[key], "soft_yield"s, value.soft_yield, path + to_string(key) + "/");
```
diegonehab, line 439: There is a matter of consistency here. Other options are "do this". This option is "this". Do we change the other ones to be "no_root_hash_check" and "no version check", or do we change this to "honor soft yield" or some such?

diegonehab (commented):
See in-line comments.

src/grpc-virtual-machine.cpp
edubart: Good catch, indeed the way I did soft yields are lost in case it yields exactly at `mcycle_end`, will fix this.

edubart: So, I did this comparing `response.mcycle() != mcycle_end` because I was trying to avoid doing changes in grpc interfaces. Now I did the proper way by forwarding from the run command, but I had to create a PR on grpc interfaces https://github.com/cartesi/grpc-interfaces/pull/28

src/interpret.cpp
edubart: Also changed here.

src/json-util.cpp
edubart: So, nothing to do here now, right? Leave at is?
This PR is not about renaming `skip` flag to `no`, you had your chance on the PR these flags were added :)

diegonehab: I guess I'll have to recruit @vfusco. :)

diegonehab (approved):

diegonehab (dismissed):

vfusco (approved):

diegonehab (approved):

PR #172, opened 2024-01-11
VirtIO implementation

edubart:
**WIP**

This PR is a rebase and continuation of https://github.com/cartesi/machine-emulator/pull/39

This PR is huge, it implements:
- PLIC device
- VirtIO Console device
- VirtIO Net device
- VirtIO 9P device
- `iunrep` and PLIC related CSRs
- Many new machine configs
- Many new command line options

I will add edit here with more details on what is being done later.


src/plic.cpp
```diff
@@ -0,0 +1,234 @@
...
+/// \brief PLIC device read callback. See ::pma_write.
+static execute_status plic_write(void *context, i_device_state_access *a, uint64_t offset, uint64_t val,
+    int log2_size) {
+    (void) context;
+#ifdef DEBUG_PLIC_MMIO
+    (void) fprintf(stderr, "plic: mmio write offset=0x%lx log2_size=%d val=0x%x\n", (long) offset, log2_size,
+        (int) val);
+#endif
+
+#if 0
```
diegonehab, line 180: Have we stopped using BBL already?

src/cartesi-machine.lua
```diff
@@ -896,7 +891,7 @@ local options = {
             flash_length.root = nil
             flash_shared.root = nil
             table.remove(flash_label_order, 1)
-            bootargs = "quiet earlycon=sbi console=hvc0"
+            bootargs = bootargs:gsub(" rootfstype=.*$", "")
```
diegonehab, line 1156: Is this option always at the end of the bootargs when the option is triggered? It has to be the first option or something like this?

src/device-state-access.h
```diff
@@ -139,11 +139,13 @@ class device_state_access : public i_device_state_access {
         return m_a.read_htif_iyield();
     }
 
-    // LCOV_EXCL_START
-    void do_write_memory(uint64_t paddr, const unsigned char *data, uint64_t log2_length) override {
-        return m_a.write_memory(paddr, data, log2_length);
+    bool do_read_memory(uint64_t paddr, unsigned char *data, uint64_t length) override {
```
diegonehab, line 142: I am trying to understand what the implication of this is for uarch. What happens when you compile the interpreter to uarch? I see the method is not implemented there in uarch-machine-state-access.h. So the idea is that, if you are using virtio, you can't run the uarch? Or at least, it wouldn't work if you did? Should we add some test for this situation?

src/os.cpp
```diff
@@ -153,21 +174,71 @@ static int get_ttyfd(void) {
...
+            return true;
+        }
+    } else {
+#ifdef DEBUG_OS
+        (void) fprintf(stderr, "os_update_tty_size(): ioctl TIOCGWINSZ failed\n");
+#endif
+    }
+
+#else
+    // TODO(edubart): what to do on Windows and MacOS?
```
diegonehab, line 206: Are we going to figure this out before merge?

src/machine.cpp
```diff
@@ -438,6 +440,16 @@ machine::machine(const machine_config &c, const machine_runtime_config &r) :
     // Register pma board shadow device
     register_pma_entry(make_shadow_pmas_pma_entry(PMA_SHADOW_PMAS_START, PMA_SHADOW_PMAS_LENGTH));
 
+    // TODO(edubart): user should be able to configure these devices
```
diegonehab: I think these got disentangled in a future commit I have not yet reached. Just registering the thought here.

src/machine.cpp
```diff
@@ -494,6 +506,11 @@ machine::machine(const machine_config &c, const machine_runtime_config &r) :
     // Sort it by increasing start address
     std::sort(m_mrds.begin(), m_mrds.end(),
         [](const machine_memory_range_descr &a, const machine_memory_range_descr &b) { return a.start < b.start; });
+
+    // Disable SIGPIPE handler, because this signal be raised and terminate the emulator process,
```
diegonehab, line 548: "can be raised"?

src/machine.cpp
```diff
@@ -520,6 +537,54 @@ machine::machine(const std::string &dir, const machine_runtime_config &r) : mach
...
+bool machine::poll_virtio_devices_after_select(int select_ret, select_fd_sets *fds, i_device_state_access *da) {
+    bool interrupt_requested = false;
+    for (auto &vdev : m_vdevs) {
+        interrupt_requested |= vdev->poll_after_select(select_ret, fds, da);
+    }
+    return interrupt_requested;
+}
+
+bool machine::poll_virtio_devices(uint64_t *timeout_us, i_device_state_access *da) {
+    struct poll_virtio_upvalues {
```
diegonehab, line 593: Why not simply use captures in your lambdas?

src/state-access.h
```diff
@@ -414,22 +414,41 @@ class state_access : public i_state_access<state_access, pma_entry> {
...
+                // VirtIO console device will poll TTY
+            } else if (has_virtio_devices) { // VirtIO without a console + HTIF console
+                m_m.poll_virtio_devices(&timeout_us, &da);
+                // Poll tty without waiting more time, because the pool above should have waited enough time
+                os_poll_tty(0);
+            } else { // Only HTIF console
+                os_poll_tty(timeout_us);
+            }
+            // If timeout is greater than zero, we should also increment mcycle relative to the elapsed time
+            if (timeout_us > 0) {
```
diegonehab, line 451: There is another option for this too, I guess.

src/state-access.h
```diff
@@ -464,8 +483,26 @@ class state_access : public i_state_access<state_access, pma_entry> {
         aliased_aligned_write(hpage + hoffset, val);
     }
 
-    void do_write_memory(uint64_t paddr, const unsigned char *data, uint64_t log2_size) {
-        m_m.write_memory(paddr, data, UINT64_C(1) << log2_size);
+    bool do_read_memory(uint64_t paddr, unsigned char *data, uint64_t length) const {
```
diegonehab, line 493: Can we assume this will work? Can this call ever be issued with an invalid data/length combination?

src/virtio-console.cpp
```diff
@@ -0,0 +1,164 @@
...
+    notify_config_change(a);
+    return true;
+}
+
+void virtio_console::poll_before_select(select_fd_sets *fds, uint64_t *timeout_us) {
+    // Ignore if driver is not initialized
+    if (!driver_ok) {
+        return;
+    }
+    // We should not poll console before the guest is actually started waiting for inputs,
```
diegonehab, line 132: guest is actually started -> guest has started
will make -> will cause

src/machine.h
```diff
@@ -288,6 +294,36 @@ class machine final {
     /// \brief Destructor.
     ~machine();
 
+    /// \brief Fill file descriptors to be polled by select() for all VirtIO devices.
+    /// \param fds Pointer to sets of read, write and except file descriptors to be updated.
+    /// \param timeout_us Maximum amount of time to wait in microseconds, this may be updated (always to lower values).
+    void poll_virtio_devices_before_select(select_fd_sets *fds, uint64_t *timeout_us);
```
diegonehab, line 301: This function name confused. It seems like this function does poll the virtio devices before some unrelated call to select. I'd just say prepare_virtio_devices_for_select.

uarch/uarch-machine-state-access.h
```diff
@@ -522,8 +522,8 @@ class uarch_machine_state_access : public i_state_access<uarch_machine_state_acc
         return raw_read_memory<uint64_t>(shadow_state_get_csr_abs_addr(shadow_state_csr::htif_iyield));
     }
 
-    uint64_t do_poll_console(uint64_t mcycle) {
-        return mcycle;
+    std::pair<uint64_t, bool> do_poll_external_interrupts(uint64_t mcycle, uint64_t mcycle_max) {
```
diegonehab, line 533: So the idea is really for uarch to operate as though virtio didn't exist?

uarch/uarch-machine-state-access.h
```diff
@@ -541,7 +541,15 @@ class uarch_machine_state_access : public i_state_access<uarch_machine_state_acc
         *pval = raw_read_memory<T>(paddr);
     }
 
-    void do_write_memory(uint64_t paddr, const unsigned char *data, uint64_t log2_size) {}
+    bool do_read_memory(uint64_t paddr, unsigned char *data, uint64_t length) {
```
diegonehab, line 552: Should it be ever used?

src/virtio-p9fs.cpp
```diff
@@ -0,0 +1,1828 @@
...
+    // Get the fid state
+    const p9_fid_state *fidp = get_fid_state(fid);
+    if (!fidp) {
+        return send_error(msg, tag, P9_EPROTO);
+    }
+    stat_t st{};
+    // Use fd when available, because its path might have been removed while fd still open
+    if (fidp->fd >= 0) {
+        // Get the attributes
+        if (fstat(fidp->fd, &st) != 0) {
```
diegonehab, line 1088: Is it even worth it to try to get this to work on Windows? Probably a pain, right?

src/virtio-serializer.h
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    template <int N>
+    bool write_value(const char pval[N]) {
+        return write_u16_string(pval, strnlen(pval, N));
+    }
+
+    template <typename T, typename... Rest>
+    bool pack(T *pval, Rest... rest) {
+        if (!write_value(pval)) {
```
diegonehab, line 178: Fold expression?

src/dtb.cpp
```diff
@@ -161,6 +161,16 @@ void dtb_init(const machine_config &c, unsigned char *dtb_start, uint64_t dtb_le
                 fdt.prop_u32_list<2>("interrupts-extended", {PLIC_PHANDLE, plic_irq_id});
                 fdt.end_node();
             }
+            if (c.htif.console_getchar) { // virtio net
```
diegonehab: Also the logic will change, right?

src/machine-config.h
```diff
@@ -122,6 +125,45 @@ struct htif_config final {
...
+
+/// \brief VirtIO Plan 9 filesystem device state config
+struct virtio_p9fs_config final {
+    std::string tag{};            ///< Guest mount tag
+    std::string host_directory{}; ///< Path to the host shared directory
+};
+
+/// \brief VirtIO host forward state config
+struct virtio_hostfwd_config final {
+    bool is_udp{false};
```
diegonehab, line 140: Is there any chance there would be another protocol ever? Say, unix? Also, what interface are you binding to on the host?

src/machine.cpp
```diff
@@ -444,28 +444,44 @@ machine::machine(const machine_config &c, const machine_runtime_config &r) :
...
-        // Register VirtIO network device
-        auto vdev_net = std::make_unique<virtio_net>(m_vdevs.size(), std::make_unique<virtio_net_carrier_slirp>());
-        register_pma_entry(
-            make_virtio_pma_entry(PMA_FIRST_VIRTIO_START + vdev_net->get_virtio_index() * PMA_VIRTIO_LENGTH,
-                PMA_VIRTIO_LENGTH, "VirtIO network device", &virtio_driver, vdev_net.get()));
-        m_vdevs.push_back(std::move(vdev_net));
+    // Initialize VirtIO devices
+    if (!m_c.virtio.empty()) {
+        // VirtIO devices are disallowed when interactive mode is disabled
+        if (!m_c.htif.console_getchar) {
```
diegonehab: I still don't think this makes much sense. There should be no connection between htif.console_getchar and virtio. They have the same "status". Did you fix this in some future commit already?

src/machine.cpp
```diff
@@ -667,6 +683,9 @@ machine_config machine::get_serialization_config(void) const {
     for (auto &f : c.flash_drive) {
         f.image_filename.clear();
     }
+    if (!c.virtio.empty()) {
```
diegonehab, line 691: Should probably check the unrep, right?

src/machine.cpp
```diff
@@ -766,6 +785,9 @@ static inline T &deref(T *t) {
 }
 
 void machine::store_pmas(const machine_config &c, const std::string &dir) const {
+    if (!c.virtio.empty()) {
```
diegonehab, line 793: Again check unrep?

src/virtio-net-carrier-slirp.cpp
```diff
@@ -197,12 +201,18 @@ virtio_net_carrier_slirp::virtio_net_carrier_slirp() {
...
-    struct in_addr host_addr {};
-    struct in_addr guest_addr {};
-    host_addr.s_addr = htonl(INADDR_ANY);
-    guest_addr.s_addr = htonl(SLIRP_DEFAULT_IPV4_VDHCP_START);
-    if (slirp_add_hostfwd(slirp, false, host_addr, 2022, guest_addr, 22) < 0) {
-        throw std::runtime_error("failed to forward host TCP port 2022 to guest port 22");
+    for (const auto &hostfwd : config.hostfwd) {
+        struct in_addr host_addr {};
+        struct in_addr guest_addr {};
+        // ??(edubart): Should we allow configuring host/guest addrs?
```
diegonehab, line 207: I think that forcing host address to be INADDR_ANY is not ideal. I commented on the config already.

src/machine.cpp
```diff
@@ -527,7 +528,10 @@ machine::machine(const machine_config &c, const machine_runtime_config &r) :
     }
 
     // Initialize TTY if console input is enabled
-    if (m_c.htif.console_getchar) {
+    if (m_c.htif.console_getchar || has_virtio_console()) {
```
diegonehab, line 531: Can you still set both htif.console_getchar and add a virtio console? Shouldn't we forbid this?

src/plic.cpp
edubart: Yes, right? And we won't support I a guess?

src/cartesi-machine.lua
```diff
@@ -292,14 +292,137 @@ where options are:
   --max-uarch-cycle=<number>
     stop at a given micro cycle.
 
+  --unreproducible
```
diegonehab, line 295: This does not automatically enable htif console getchar nor virtio console, right?

src/os.cpp
edubart: There are still some `TODO(edubart)`, the PR is WIP, they will all be fixed before merging.

src/machine.cpp
edubart: Indeed, I forgot to remove the line.

src/machine.cpp
edubart: To not trigger allocation, lambda captures will allocate memory.

diegonehab (changes requested):
See comments.

src/plic.cpp
diegonehab: It's pretty obsolete, right? Should we just sunset it?

src/machine.cpp
diegonehab: Why would a capture allocate memory? In the stack, you mean?

src/cartesi-machine.lua
edubart: Correct, this option by itself will not trigger any console or terminal related polling.

src/virtio-net-carrier-slirp.cpp
edubart: Should I use `INADDR_LOOPBACK` instead? I am trying to avoid this to be configurable.

src/machine.cpp
edubart: You commented in an old commit, this was changed to check `iunrep` instead, although I forgot to change the comment.

src/machine-config.h
edubart: > What interface are you binding to on the host.
`libslirp` has no option to choose an interface, I guess this can be automatically picked by the host address the machine is binding, I am thinking in binding always to loopback, so `lo` interface, do you agree?

src/dtb.cpp
edubart: You commented in an outdated commit, the logic was changed.

src/virtio-p9fs.cpp
edubart: I want to at least allow to compile the cartesi machine on Windows / WASM, with some VirtIO being obviously unsupported. This is something I have to do yet with some ifdefs and makefile rules.

src/dtb.cpp
```diff
@@ -151,6 +151,16 @@ void dtb_init(const machine_config &c, unsigned char *dtb_start, uint64_t dtb_le
                 fdt.prop_u32_list<2>("interrupts-extended", {PLIC_PHANDLE, plic_irq_id});
                 fdt.end_node();
             }
+            if (c.htif.console_getchar) { // virtio 9p
```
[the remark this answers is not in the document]
edubart: You commented in an outdated commit, the logic was changed.

src/machine.cpp
edubart: I was under the impression that non trivial lambdas (with captures) could cause heap allocation when converted to an `std::function` at the moment, so I avoided them, this is not that obvious. But after some research seems like C++ is smart enough to avoid triggering heap allocation.

src/state-access.h
```diff
@@ -414,22 +414,41 @@ class state_access : public i_state_access<state_access, pma_entry> {
...
+            if (timeout_us > 0) {
+                start_us = os_now_us();
+            }
+            device_state_access da(*this, mcycle);
+            // Poll virtio for events (e.g console stdin, network sockets)
+            // Timeout may be decremented in case a device has deadline timers (e.g network device)
+            if (has_virtio_devices && has_virtio_console) { // VirtIO + VirtIO console
+                m_m.poll_virtio_devices(&timeout_us, &da);
+                // VirtIO console device will poll TTY
+            } else if (has_virtio_devices) { // VirtIO without a console + HTIF console
```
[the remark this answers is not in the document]
edubart: Having a network, but no console is indeed a thing to run scripts. This is something I tested and should work.

The logic changed a bit in the next commits. But it is fine calling `os_poll_tty` even when console is disabled, because it will return immediately since TTY was not initialized.

uarch/uarch-machine-state-access.h
edubart: Correct, the idea is to not ever support any VirtIO device in uarch.

src/machine.cpp
edubart: Yes you can set both. I don't think we should forbid, so if the OS fails to initialize VirtIO it can fallback to HTIF console if enabled. I don't see issues having both enabled, this is something I did for a while.

uarch/uarch-machine-state-access.h
edubart: This is not triggered in uarch, just needed to compile.

src/cartesi-machine.lua
edubart: All options after `rootfstype=` are related to the rootfs, the idea here is to clear out all rootfs ralted options from bootargs.

diegonehab: These options are the absolute last thing you add to the bootargs, right? We should probably add a comment that this *has* to be the case, then.

src/device-state-access.h
diegonehab: So the interpreter that is for uarch, when it is executing a bit SD, and searching PMAs, doesn't find the PMA for the virtio devices?

src/state-access.h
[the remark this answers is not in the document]
diegonehab: It's just for clarity.

uarch/uarch-machine-state-access.h
diegonehab: We should probably check for the nrep and bail out from uarch with an error? Just to be safe?

src/virtio-p9fs.cpp
diegonehab: What does qemu do? (Not saying we should do the same. Just wondering.)

src/virtio-serializer.h
diegonehab: C++17 introduces fold expressions with the "comma" operator to simplify variadic templates. You almost never need recursion anymore.

diegonehab: It even strengthened the expression evaluation order guarantees of C++ to make sure these expressions are well defined. A thing of beauty.

src/device-state-access.h
edubart: > So the interpreter that is for uarch, when it is executing a bit SD, and searching PMAs, doesn't find the PMA for the virtio devices?

Correct, VirtIO was not added to `build_uarch_pma_entry` in `uarch-machine-state-access.h`, SD will fail. But now I  added comments + `abort()` these places (`do_write_memory`, `do_read_memory`, `build_uarch_pma_entry`).

Plus I added a runtime error when trying to perform `uarch_run` with `iunrep` set, which should cover VirtIO being enabled.

src/machine.cpp
edubart: I converted to lambdas with stack captures.

src/state-access.h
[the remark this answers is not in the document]
edubart: Added some new comments here, with more clarifications of all possible scenarios.

src/state-access.h
edubart: I think before VirtIO this would never fail, because no device was performing memory writes on addresses choosing by guest drivers, but now this is a thing with VirtIO. 

VirtIO guest drivers will pass physical memory addresses for VirtIO devices buffers, and if the driver is mess up these addresses, this must return false so the driver can recover up and reset.

So yes, this function can be issued with invalid data/length combinations from VirtIO devices.

src/machine.h
edubart: Changed the names.

src/machine.h
```diff
@@ -288,6 +294,36 @@ class machine final {
...
+
+    /// \brief Poll file descriptors that were marked as ready by select() for all VirtIO devices.
+    /// \param select_ret Return value from the most recent select() call.
+    /// \param fds Pointer to sets of read, write and except file descriptors to be checked.
+    /// \returns True if an interrupt was requested, false otherwise.
+    /// \details This function process pending events and trigger interrupt requests (if any).
+    bool poll_virtio_devices_after_select(int select_ret, select_fd_sets *fds, i_device_state_access *da);
```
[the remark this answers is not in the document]
edubart: Changed the names.

src/virtio-serializer.h
edubart: Cool, I was not aware, I made use of it to simplify the code. I guess I am still stuck in C++11 :sweat_smile:

src/machine-config.h
edubart: > Is there any chance there would be another protocol ever?
No, libslirp support only forwarding TCP or UDP.

With my recent changes it will bind to 127.0.0.1, therefore the loopback interface.

diegonehab: I meant that, usually, people let you pick the interface to bind by specifying the associated ip address.

src/device-state-access.h
diegonehab: Great. I now feel safe. :)

src/state-access.h
diegonehab: So I guess we either replicate the PMA search logic there, or we deal with exceptions...

src/virtio-p9fs.cpp
edubart: QEMU has an abstraction layer for cross platform file operations. For now I am avoiding this problem, and Windows has no support for VirtIO 9PFS, later if there is demand we can add back.

src/machine-config.h
edubart: I changed a bit how port forwarding is done, now it is the `--port-forward` or `-p` option, it has a syntax similar to docker `-p` option and now supports setting the host address to bind to.

src/virtio-net-carrier-slirp.cpp
edubart: This is now being done and the comment was removed.

vfusco (approved):

diegonehab (approved):

PR #180, opened 2024-02-01
Refactor uarch generated files for easy cross compilation

edubart:
This PR refactors how uarch related files are generated to make it able to cross compiling for WASM and Windows again.

Now when you type `make uarch-with-linux-env` it will build the following files:
- `uarch/uarch-rom.bin`
- `uarch/uarch-pristine-hash.c`
- `uarch/uarch-pristine-ram.c`

This will be built inside the uarch Docker. Then when you do `make -C src` only the files `uarch/uarch-pristine-hash.c` and `uarch/uarch-pristine-ram.c` will be used from the `uarch` directory.
For the ones wanting to compile without Docker, downloading those two files should be enough.

Notice that if any source code in the emulator interpreter is changed, to make sure `uarch-pristine-ram.c` is correct, you have to explicit build it again by removing it or invoking `make uarch-with-linux-env`, because make timestamp dependency tree is not propagated from `src` to `uarch` directory, this is intentional to allow cross compiling.


mpernambuco (approved):
ok

vfusco (approved):

uarch/compute-uarch-pristine-hash.cpp
```diff
@@ -113,11 +113,24 @@ int main(int argc, char *argv[]) try {
     proof_type proof = tree.get_proof(UARCH_STATE_START_ADDRESS, UARCH_STATE_LOG2_SIZE, nullptr);
     auto &uarch_state_hash = proof.get_target_hash();
 
+    // Print header
```
mpolitzer, line 116: Does this change remove the `xxd` requirement?

edubart: No, this file generates only `uarch-pristine-hash.c`, but `uarch-pristine-ram.c` still needs `xxd` and this PR still includes your xxd fixes.

vfusco (approved):

mpernambuco (approved):

Issue #202, opened 2024-03-19
Jsonrpc remote machine keeps listening forever when advancing with multiple cli commands


edubart:
## Context

Test case:
```sh
# start remote cartesi machine
killall jsonrpc-remote-cartesi-machine
jsonrpc-remote-cartesi-machine --server-address=127.0.0.1:8000 --log-level=info &

# create inputs
echo '{"payload":"hello"}' | rollup-memory-range encode input > epoch-0-input-0.bin
echo '{"payload":"hello"}' | rollup-memory-range encode input > epoch-0-input-1.bin
echo '{"msg_sender":"0x7D2c4B415d9F917d6effC7761EAB3849B5EcbF39", "epoch_index": 0, "input_index": 0, "block_number":0, "time_stamp": 0}' | rollup-memory-range encode input-metadata > epoch-0-input-metadata-0.bin
echo '{"msg_sender":"0x7D2c4B415d9F917d6effC7761EAB3849B5EcbF39", "epoch_index": 0, "input_index": 1, "block_number":0, "time_stamp": 0}' | rollup-memory-range encode input-metadata > epoch-0-input-metadata-1.bin

# boot machine
cartesi-machine \
    --remote-address=127.0.0.1:8000 \
    --no-remote-destroy \
    -- rollup-init /usr/bin/echo-dapp --reports=1

# perform first advance state
cartesi-machine \
    --remote-address=127.0.0.1:8000 \
    --no-remote-create \
    --rollup-advance-state=epoch_index:0,input_index_begin:0,input_index_end:1

# perform second advance state
cartesi-machine \
    --remote-address=127.0.0.1:8000 \
    --no-remote-create \
    --rollup-advance-state=epoch_index:0,input_index_begin:1,input_index_end:2

# check listening ports
netstat -antp | grep jsonrpc
```

The final output is:
```
tcp        0      0 127.0.0.1:43599         0.0.0.0:*               LISTEN      1415841/jsonrpc-rem 
tcp        0      0 127.0.0.1:39455         0.0.0.0:*               LISTEN      1415843/jsonrpc-rem 
tcp        0      0 127.0.0.1:8000          0.0.0.0:*               LISTEN      1415830/jsonrpc-rem
```

Notice that we have 3 port listening, where just 1 was expected. If we keep doing advances like this, eventually it will exhaust all available ports on the host (about 65535). 

## Expected behavior

[...]

PR #204, opened 2024-03-19
Rework machine snapshots API

edubart:
This is a fix for https://github.com/cartesi/machine-emulator/issues/202 and also adds new command line options `--no-rollback` that fulfills https://github.com/cartesi/machine-emulator/issues/206

### DIFF
--- a/src/cartesi-machine.lua
+++ b/src/cartesi-machine.lua
@@ -2079,7 +2079,7 @@ if config.processor.iunrep ~= 0 then stderr("Running in unreproducible mode!\n")
 if store_config == stderr then store_machine_config(config, stderr) end
 if cmio_advance or cmio_inspect then
     check_cmio_htif_config(config.htif)
-    assert(remote_address, "cmio requires --remote-address for snapshot/rollback")
+    assert(remote_address, "cmio requires --remote-address for snapshot/commit/rollback")
 end
 local cycles = machine:read_mcycle()
 if initial_hash then
@@ -2135,6 +2135,7 @@ while math.ult(cycles, max_mcycle) do
         elseif cmio_advance and cmio_advance.next_input_index < cmio_advance.input_index_end then
             -- previous reason was an accept
             if reason == cartesi.machine.HTIF_YIELD_MANUAL_REASON_RX_ACCEPTED then
+                machine:commit()
                 -- save only if we have already run an input and have just accepted it
                 if cmio_advance.next_input_index > cmio_advance.input_index_begin then
                     assert(length == 32, "expected root hash in tx buffer")
--- a/src/cartesi-machine.lua
+++ b/src/cartesi-machine.lua
@@ -51,6 +51,9 @@ where options are:
     use a remote cartesi machine listening to <address> instead of
     running a local cartesi machine.
 
+  --remote-fork
+    fork the remote cartesi machine before the execution.
+
   --remote-shutdown
     shutdown the remote cartesi machine after the execution.
 
@@ -60,6 +63,13 @@ where options are:
 
   --no-remote-destroy
     do not destroy the cartesi machine in the remote server after the execution.
+
+  --no-rollback
+    disable rollback for advance and inspect states.
+    this allows to perform advance and inspect states on local cartesi machines,
+    however the state is never reverted, even in case inspects or rejected advances.
+
+    DON'T USE THIS OPTION IN PRODUCTION
 
   --ram-image=<filename>
     name of file containing RAM image (default: "linux.bin").
@@ -537,9 +547,11 @@ end
 local remote
 local remote_protocol = "jsonrpc"
 local remote_address
+local remote_fork = false
 local remote_shutdown = false
 local remote_create = true
 local remote_destroy = true
+local perform_rollbacks = true
 local images_path = adjust_images_path(os.getenv("CARTESI_IMAGES_PATH"))
 local flash_image_filename = { root = images_path .. "rootfs.ext2" }
 local flash_label_order = { "root" }
@@ -1292,6 +1304,14 @@ local options = {
         end,
     },
     {
+        "^%-%-remote%-fork$",
+        function(o)
+            if not o then return false end
+            remote_fork = true
+            return true
+        end,
+    },
+    {
         "^%-%-remote%-shutdown$",
         function(o)
             if not o then return false end
@@ -1312,6 +1332,14 @@ local options = {
         function(o)
             if not o then return false end
             remote_destroy = false
+            return true
+        end,
+    },
+    {
+        "^%-%-no%-rollback$",
+        function(o)
+            if not o then return false end
+            perform_rollbacks = false
             return true
         end,
     },
@@ -1745,6 +1773,7 @@ if remote_address then
     remote = assert(protocol.stub(remote_address))
     local v = assert(remote.get_version())
     stderr("Connected: remote version is %d.%d.%d\n", v.major, v.minor, v.patch)
+    if remote_fork then remote = assert(protocol.stub(remote.fork())) end
     local shutdown = function() remote.shutdown() end
     if remote_shutdown then
         setmetatable(remote_shutdown_deleter, {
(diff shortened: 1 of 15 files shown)


src/jsonrpc-virtual-machine.cpp
```diff
@@ -171,6 +171,15 @@ jsonrpc_mg_mgr::jsonrpc_mg_mgr(std::string remote_address) {
 }
 
 jsonrpc_mg_mgr::~jsonrpc_mg_mgr() {
```
diegonehab, line 173: Do you think this is a good idea? I am afraid somebody will use --max-mcycle and this commit will happen. Not sure this is a problem. But it some hidden behavior.

src/jsonrpc-virtual-machine.cpp
```diff
@@ -197,28 +206,37 @@ const std::string &jsonrpc_mg_mgr::get_remote_parent_address(void) const {
...
+void jsonrpc_mg_mgr::commit() {
+    // If we are not forked, there is no pending snapshot to discard, therefore we are already commited
+    if (!is_forked()) {
+        return;
+    }
+
+    // To commit, we kill the parent server and replace its address with the child's
+    bool result = false;
+    jsonrpc_request(get_mgr(), get_remote_parent_address(), "shutdown", std::tie(), result);
+    std::swap(m_address[0], m_address[1]);
```
diegonehab, line 231: Here you would call rebind, right?

src/jsonrpc-discover.json
```diff
@@ -34,6 +34,27 @@
       }
     },
 
+    {
+      "name": "rebind",
+      "summary": "Rebinds the server",
+      "params": [ {
+          "name":"port",
```
diegonehab, line 41: I think this should be an address, just like what the call to fork returns.

src/jsonrpc-virtual-machine.cpp
edubart: Without this, we would to explicit do a `commit()` call in cartesi machine before exiting the script to kill the listening JSON-RPC socket. Also people who forgot this commit() call would have zombie forks that were not shutdown.

I could remove this, as long we document that forks could become zombies in case a commit/rollback is not executed, and we explicit call `commit()` to "close" a snapshot. Maybe I could even use to-be-closed variable on Lua to make sure a commit will be called, even in case of Lua errors.

src/jsonrpc-virtual-machine.cpp
edubart: Yes, the call to rebind is just in the lines below, maybe you missed because this was done in a different commit.

src/jsonrpc-discover.json
edubart: Ok, I changed the API to use `address` argument, just like the one returned by fork.

src/jsonrpc-virtual-machine.cpp
diegonehab: I would rather we not do this automatically. I think the right place for something like this is in cartesi-machine.lua. There, you can create an object to-be-closed with a metamethod that does this, after checking a command-line flag that lets the user disable the behavior.

diegonehab (changes requested):
See request for change.

src/jsonrpc-virtual-machine.cpp
edubart: I removed this code, now the `cartesi-machine.lua` will clean this. I didn't feel the need to introduce a new command-line option to disable this behavior, meaning snapshots will be committed only when an advance state is accepted, all other cases will cause rollback in the remote machine before the cli exits, such as:
- Any raised lua error
- A machine that is stopped before an advance state is completed (due to max mcycle)
- Any inspects

However I took the opportunity to introduce a new option `--no-rollback`, to allow performing advance/inspect states on a local machine, with is a BIG disclaimer that state will never be reverted. This makes the cli a little more convenient to perform tests of advances/inspects without the need of spawning remote cartesi machines.

src/cartesi-machine.lua
```diff
@@ -2159,7 +2215,7 @@ while math.ult(cycles, max_mcycle) do
                 if cmio_inspect.query then
                     stderr("\nBefore query\n")
                     if cmio_inspect.hashes then print_root_hash(machine) end
-                    machine:snapshot()
+                    do_snapshot(machine)
```
mpolitzer, line 2218: ?

```suggestion
                    do_snapshot()
```

src/jsonrpc-discover.json
```diff
@@ -34,6 +34,27 @@
       }
     },
 
+    {
+      "name": "rebind",
+      "summary": "Rebinds the server",
```
diegonehab, line 39: "Changes the address server is listening to."

Most people will have no idea what bind even is. :)

src/jsonrpc-discover.json
```diff
@@ -34,6 +34,27 @@
...
+      "params": [ {
+          "name": "address",
+          "description": "URL to rebind",
```
diegonehab, line 42: URL of new address

src/jsonrpc-machine-c-api.h
```diff
@@ -152,6 +152,15 @@ CM_API int cm_jsonrpc_verify_uarch_step_state_transition(const cm_jsonrpc_mg_mgr
 /// \returns 0 for successful verification, non zero code for error
 CM_API int cm_jsonrpc_fork(const cm_jsonrpc_mg_mgr *mgr, char **address, char **err_msg);
 
+/// \brief Rebinds the server
```
diegonehab, line 155: "Changes the address server is listening to."

src/jsonrpc-virtual-machine.cpp
```diff
@@ -223,11 +223,18 @@ void jsonrpc_mg_mgr::commit() {
         return;
     }
 
+    const std::string parent_address = m_address[0];
+
     // To commit, we kill the parent server and replace its address with the child's
     bool result = false;
     jsonrpc_request(get_mgr(), get_remote_parent_address(), "shutdown", std::tie(), result);
     std::swap(m_address[0], m_address[1]);
```
diegonehab, line 222: Isn't this a pretty complicated logic for what should simply be a pop_back? Maybe I misread...

src/cartesi-machine.lua
```diff
@@ -2074,6 +2103,36 @@ if periodic_hashes_start ~= 0 then
 else
     next_hash_mcycle = periodic_hashes_period
 end
+
+-- proxy functions to snapshot/commit/rollback
+local forked_snapshot = false
+local function do_snapshot()
```
diegonehab, line 2109: I would simply call these snapshot/commit/rollback and pass the machine to them.
I think we are starting to use way too many global variables (e.g. cycle) that are modified inside functions. Can we try to minimize this?

diegonehab (changes requested):
See comments in code.

src/jsonrpc-virtual-machine.cpp
edubart: This is exactly the same logic as before, I just added the few lines below. I did not simplify in case the next rebind request throws.

src/cartesi-machine.lua
edubart: I was passing before but the `machine` word is already in use and was giving me Lua warnings, so I decided to remove, also because it will depend on the global `forked_snapshot` anyway.

src/jsonrpc-virtual-machine.cpp
diegonehab: I understand. But if the rebind throws, it's difficult to recover anyway, right? It may or may not have changed address before throwing. So we now need to figure out which one we should use to contact it. I think the only way is to try both?

edubart: I simplified and just left a pop_back(). Continue to use the connection after a throw here is undefined behavior :) and the caller should give up on this connection, moreover the remote server could left zombie servers.

src/cartesi-machine.lua
edubart: Reworked this, passing `machine` now, and I also removed `cycles` variable in favor of `machine:read_mcycle()`.

src/cartesi-machine.lua
```diff
@@ -2197,7 +2255,19 @@ while math.ult(cycles, max_mcycle) do
         end
         -- otherwise ignore
     end
-    if machine:read_iflags_Y() then break end
+    if machine:read_iflags_Y() then
+        -- commit/rollback depends on the status of last yield request
+        local _, reason, length = get_yield(machine)
+        if reason == cartesi.machine.HTIF_YIELD_MANUAL_REASON_TX_EXCEPTION then
```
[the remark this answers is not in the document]
edubart: Removed the treatment here, and added a early rollback and break in the TX_EXCEPTION handling above.

diegonehab (approved):

vfusco (approved):

PR #218, opened 2024-03-29
bugfix: EADDRNOTAVAIL on rapid successive requests

mpernambuco:


src/jsonrpc-virtual-machine.cpp
```diff
@@ -57,10 +57,25 @@ struct http_request_data {
...
+// Performs additional client socket configuration
+static void setup_client_socket(struct mg_connection *c) {
+    // Minimize socket close time: by setting the linger time to 0.
+    // Avoid accumulating socket in TIME_WAIT state in rapid successive requests, which can consume all available ports.
+    // It is safe to do this because this client decides to close the connection, after all data is received.
+    struct linger so_linger {};
+    so_linger.l_onoff = 1;
+    so_linger.l_linger = 0; // do not wait
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    auto socket = static_cast<MG_SOCKET_TYPE>(reinterpret_cast<size_t>(c->fd));
```
diegonehab, line 69: A socket is an  regular signed ‘int’ on Unix and an unsigned `SOCKET` on Windows. I don’t know what mongoose does when it casts the damn thing to a pointer type. INVALID_SOCKET on windows is a large positive number. So converting to `size_t` is ok. But converting -1 to a pointer than to a `size_t` than back to `int` will probably overflow in an undefined way. So I dunno.

diegonehab (approved):

src/jsonrpc-virtual-machine.cpp
```diff
@@ -57,10 +57,25 @@ struct http_request_data {
...
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    (void) setsockopt(socket, SOL_SOCKET, SO_LINGER, reinterpret_cast<char *>(&so_linger), sizeof(so_linger));
```
edubart, line 71: I suggest to wrap this function contents inside an `#if defined(SO_LINGER)` so we can still compile this file on platforms that does not support this socket option.

edubart (commented):
Made a small comment on how to make the code more portable across different platforms.

diegonehab (approved):

edubart (approved):

PR #220, opened 2024-04-01
feat: add C API to translate virtual address

edubart:
This PR adds a new function to the public C API:

1. `cm_translate_virtual_address` - Translates a virtual memory address to its corresponding physical memory address

The use case is to debug address translation from outside, this function was exposed to Lua and JSONRPC.

### DIFF
--- a/src/clua-i-virtual-machine.cpp
+++ b/src/clua-i-virtual-machine.cpp
@@ -499,6 +499,16 @@ static int machine_obj_index_write_virtual_memory(lua_State *L) {
     return 1;
 }
 
+/// \brief This is the machine:translate_virtual_address() method implementation.
+/// \param L Lua state.
+static int machine_obj_index_translate_virtual_address(lua_State *L) {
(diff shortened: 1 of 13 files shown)


mpolitzer:
I would be more explicit with the naming: `translate_guest_physical_to_host_ptr` (may be a bit overkill, but you get the idea).
Also this only makes sense when looking up within page boundaries, do you plan to use this with large pages?

edubart:
> Also this only makes sense when looking up within page boundaries, do you plan to use this with large pages?
I do plan to use this on personal projects, taking care to use the added methods carefully, always within page boundaries.

> `translate_guest_physical_to_host_ptr`
Yea, kinda overkill, @diegonehab do you have name suggestions?

diegonehab (changes requested):
Can we change this to "translate_virtual_address"? You are not really translating the virtual memory, after all.

diegonehab (approved):

PR #237, opened 2024-04-29
Add support for snapshot/rollback in local machines

edubart:
This is an experimental PR adding support for performing snapshot/rollback on local machines. Making the APIs available in Windows and WASM. It uses the naive implementation of duplicating the whole machine memory.

This PR does not try to optimize this operation to compete with JSONRPC `fork()`, it just provides the naive implementation that will work on any system.

## TODO
- [ ] add lua tests
- [ ] make deep copy of merkle tree and use it
- [ ] avoid deep copy of read only PMAs and design a way to share them between multiple machines
  - [ ] introduce read only flash drives, so we can avoid copying them, useful for machines with read-only rootfs
- [ ] optimize Windows memory map to not use calloc
- [ ] add disable CMIO option again, so we don't have to deep copy them on WASM
- [ ] add fork method to machine C API?

## Benchmark

To comparise the of performance of `local` vs `jsonrpc` machines, I tested booting the last release of tools rootfs tanking random roollbacks:

```lua
math.randomseed(0)
while not machine:read_iflags_H() do
    machine:snapshot()
    machine:run(machine:read_mcycle() + 1000)
    if math.random() < 0.3 then
        machine:rollback()
    else
        machine:commit()
    end
end
```

This is the result:
```
local       67 iterations/s
jsonrpc    895 iterations/s
```

Therefore `jsonrpc` is about 13x times faster than `local`, the performance for local would be even worse from bigger machines.

### DIFF
--- a/src/cartesi-machine.lua
+++ b/src/cartesi-machine.lua
@@ -2106,10 +2106,7 @@ if gdb_address then
 end
 if config.processor.iunrep ~= 0 then stderr("Running in unreproducible mode!\n") end
 if store_config == stderr then store_machine_config(config, stderr) end
-if cmio_advance or cmio_inspect then
-    check_cmio_htif_config(config.htif)
-    assert(remote_address or not perform_rollbacks, "cmio requires --remote-address for snapshot/commit/rollback")
-end
+if cmio_advance or cmio_inspect then check_cmio_htif_config(config.htif) end
 if initial_hash then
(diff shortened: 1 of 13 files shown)


edubart:
This PR is being superseded by https://github.com/cartesi/machine-emulator/pull/281 , which will be superior and faster.

PR #239, opened 2024-05-13
fix: update makefile to make it compatible with a homebrew formulae

alexmikhalevich:
This fix makes our Makefile compatible with Homebrew formulae. Below is an explanation of the patch:

1. During the processing of a formula by Homebrew, `$(brew --prefix)/bin` is not included in the `PATH`, resulting in an empty result for which brew. To address this issue, I propose allowing manual setting of `BREW_PREFIX`. This manual approach will be utilized in the formula: `system "make", "BREW_PREFIX=#{prefix}"`. The flow of `PORT_PREFIX` was changed for consistency.
1. During the execution of `make install`, [a couple of targets from src/Makefile](https://github.com/cartesi/machine-emulator/blob/1837a0ce12997eec802460662fcb0c3ae991ac0a/Makefile#L23) are used for version generation. When make install is invoked from the context of Homebrew formula installation, `BREW_PREFIX` is empty, causing the installation process to abort with the message `Neither Homebrew nor MacPorts is installed`. I propose changing error to warning here. This will prevent the process from aborting, making the user responsible for not specifying a prefix. Considering that this aspect is intended for developers only, I believe this approach is adequate.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -66,27 +66,30 @@ CXX=clang++
 AR=libtool -static -o
 INCS=
 
+BREW_PREFIX=$(shell which brew)
+PORT_PREFIX=$(shell which port)
+
 ifeq ($(MACOSX_DEPLOYMENT_TARGET),)
 export MACOSX_DEPLOYMENT_TARGET := $(shell sw_vers -productVersion | sed -E "s/([[:digit:]]+)\.([[:digit:]]+)\..+/\1.\2.0/")
 endif
 
 # Homebrew installation
-ifneq (,$(shell which brew))
-BREW_PREFIX = $(shell brew --prefix)
+ifneq (,$(BREW_PREFIX))
 BOOST_LIB_DIR=-L$(BREW_PREFIX)/lib
 BOOST_INC=-I$(BREW_PREFIX)/include
 SLIRP_LIB=-L$(BREW_PREFIX)/lib -lslirp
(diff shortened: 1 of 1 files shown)


diegonehab (approved):

src/Makefile
```diff
@@ -66,27 +66,29 @@ CXX=clang++
 AR=libtool -static -o
 INCS=
 
+BREW_PREFIX=$(shell which brew)
+PORT_PREFIX=$(shell which port)
```
edubart, line 70: I am not on MacOS to test, but before PORT_PREFIX was `/opt/local` now will become path to the `port` binary, I don't know where it is, but I imagine it would be `/opt/bin/port`. What `shell which port` actually returns? Is it `/opt/local` ?

mpolitzer (approved):

src/Makefile
diegonehab: Why are we calling it PREFIX if it is the path to the package manager executable?

edubart: Ah I got, so the wanted behavior was really the executable path, `_PREFIX` tricked me thinking it would need to point to a directory, not an executable.

diegonehab: It's just that "prefix" is something in "configure" and "makefile" jargon that is supposed to mean the base directory on top of which you will end up installing the products of the source tree. Hence Eduardo's comment.

edubart (approved):

Issue #222, opened 2024-04-02
Optimize jsonrpc request throughput


edubart:
## Context

Currently these are some numbers of very lightweight advance requests (without forks) made through local/jsonrpc/grpc machines:

```
local 2819.05 req/s
jsonrpc 491.62 req/s
grpc 69.71 req/s
```

Local is way faster than jsonrpc, jsonrpc adds the overhead of:
- Networking
- Establishing a new connection for every machine method call
- JSON serialization/deserialization

## Possible solutions

We can optimize jsonrpc requests during a remote machine session by using keep alive connections, if that is not enough, we should also investigate if there are simple ways to optimize the serialization or networking.

Issue #157, opened 2023-11-07
`make dep` fails if `third-party/downloads/` directory exists (even if it is empty)


mpolitzer:
## Context

> What is the bug that you are experiencing?

Rebuilding the emulator fails with certain conditions.

> Why is this bug relevant?

This is relevant during the development process and requires manual intervention besides `make` calls

## Expected behavior

`make dep` should download the file: `third-party/downloads/7.12.tar.gz`

## Actual behavior

```
make dep                                                                                                                                                               
tar -C third-party -xzf third-party/downloads/7.12.tar.gz mongoose-7.12/mongoose.c mongoose-7.12/mongoose.h                                                                                                             
tar (child): third-party/downloads/7.12.tar.gz: Cannot open: No such file or directory                                                                                                                                  
tar (child): Error is not recoverable: exiting now                                                                                                                                                                      
tar: Child returned status 2                                                                                                                                                                                            
tar: Error is not recoverable: exiting now                                                                                                                                                                              
make: *** [Makefile:171: third-party/mongoose-7.12] Error 2                                                                                                                                                             
```

## Steps to reproduce

create the `third-party/downloads/` directory before `make dep`

## Environment

Please describe the environment where the bug happens.

Include the following (when applicable and/or relevant):
- Emulator version: [latest] 88bf5749df883569ae361b4f4e66851259f0a0d7

## Possible solutions

Is there any workaround for the issue? (This may help others experiencing the same issue.)
Manually removing the `third-party/downloads/` directory solved the issue.

Can you offer any hint on how to solve the issue?
`make depclean` should remove the `third-party/downloads/` directory

edubart:
`make dep` and `make downloads` will go away with https://github.com/cartesi/machine-emulator/pull/240

PR #240, opened 2024-05-13
Replace mongoose with Boost ASIO and support keep alive connections

edubart:
Implementation details:

- The client uses synchronous IO.
- In the client, special care is taken when connect/write/read calls are interrupted, so the client does not lose track of spawned remote machines, otherwise it could leave dead zombies when its process receives SIGTERM/SIGINT.
- The client persist connections with keep alive, so we speed up request throughput
- The client still use SO_LINGER option to minimize TIME_WAIT connections during fork
- Both client and server uses TCP_NODELAY to minimize connection latency
- The server uses asynchronous IO, so we can supporting connecting multiple clients in parallel and forking them.
- The server only shut downs when all asynchronous operations are either completed or cancelled. Meaning SIGTERM will not make it close right-away, it will only gracefully terminate when all clients sessions are gracefully closed by the client.
- In the server, special care is taken during the fork.
- Server uses SO_REUSEADDR flag
- I've opted to disable threads, EPOLL and EVENTFD, to force Boost ASIO use `select()` based implementation, so we have less chance of issues with `fork()`.
- Boost ASIO was used for networking
- Boost Beast was used for parsing HTTP requests/responses
- There is no need to link any library, only header-only libraries from Boost were used.
- Nlohmann JSON library was bundled in third party
- `make dep` and `make downloads` were  removed
- I did some stress tests on the changes, ran thousands of requests/forks per second, found no issue so far.
- All mongoose related code was removed.

This replaces:
- https://github.com/cartesi/machine-emulator/pull/236

This should close:
- #222
- #157

### DIFF
--- a/src/jsonrpc-mgr.h
+++ b/src/jsonrpc-mgr.h
@@ -14,31 +14,37 @@
 // with this program (see COPYING). If not, see <https://www.gnu.org/licenses/>.
 //
 
-#ifndef JSONRPC_MG_MGR_H
-#define JSONRPC_MG_MGR_H
+#ifndef JSONRPC_MGR_H
+#define JSONRPC_MGR_H
 
-#include <boost/container/static_vector.hpp>
-#include <mongoose.h>
 #include <string>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#include "asio-config.h" // must be included before any ASIO header
+#include <boost/asio/io_context.hpp>
+#include <boost/beast/core/tcp_stream.hpp>
+#include <boost/container/static_vector.hpp>
+#pragma GCC diagnostic pop
 
 namespace cartesi {
 
-class jsonrpc_mg_mgr final {
-
+class jsonrpc_mgr final {
+    boost::asio::io_context m_ioc{1};         // The io_context is required for all I/O
+    boost::beast::tcp_stream m_stream{m_ioc}; // TCP stream for keep alive connections
     boost::container::static_vector<std::string, 2> m_address{};
-    struct mg_mgr m_mgr {}; // unnecessary initialization to silince clang-tidy
 
 public:
-    explicit jsonrpc_mg_mgr(std::string remote_address);
-    jsonrpc_mg_mgr(const jsonrpc_mg_mgr &other) = delete;
-    jsonrpc_mg_mgr(jsonrpc_mg_mgr &&other) noexcept = delete;
-    jsonrpc_mg_mgr &operator=(const jsonrpc_mg_mgr &other) = delete;
-    jsonrpc_mg_mgr &operator=(jsonrpc_mg_mgr &&other) noexcept = delete;
-    ~jsonrpc_mg_mgr();
+    explicit jsonrpc_mgr(std::string remote_address);
+    jsonrpc_mgr(const jsonrpc_mgr &other) = delete;
+    jsonrpc_mgr(jsonrpc_mgr &&other) noexcept = delete;
+    jsonrpc_mgr &operator=(const jsonrpc_mgr &other) = delete;
+    jsonrpc_mgr &operator=(jsonrpc_mgr &&other) noexcept = delete;
+    ~jsonrpc_mgr();
     bool is_forked(void) const;
     bool is_shutdown(void) const;
-    struct mg_mgr &get_mgr(void);
-    const struct mg_mgr &get_mgr(void) const;
+    boost::beast::tcp_stream &get_stream(void);
+    const boost::beast::tcp_stream &get_stream(void) const;
     const std::string &get_remote_address(void) const;
     const std::string &get_remote_parent_address(void) const;
     void snapshot(void);
--- a/src/jsonrpc-remote-machine.cpp
+++ b/src/jsonrpc-remote-machine.cpp
@@ -658,6 +658,7 @@ static json jsonrpc_fork_handler(const json &j, mg_connection *con, http_handler
     mg_mgr_free_ours(&h->child->event_manager);
     delete h->child;
     h->child = nullptr;
+    con->is_draining = 1;
     return jsonrpc_response_ok(j, new_server_address);
 }
 
@@ -669,7 +670,6 @@ static json jsonrpc_fork_handler(const json &j, mg_connection *con, http_handler
 /// \details Changes the address the server is listening to.
 /// After this call, all new connections should be established using the new server address.
 static json jsonrpc_rebind_handler(const json &j, mg_connection *con, http_handler_data *h) {
-    (void) con;
     static const char *param_name[] = {"address"};
     auto args = parse_args<std::string>(j, param_name);
     const std::string new_server_address = std::get<0>(args);
@@ -678,6 +678,8 @@ static json jsonrpc_rebind_handler(const json &j, mg_connection *con, http_handl
     if (!new_listen_connection) {
         return jsonrpc_response_server_error(j, "rebind failed listening on "s + new_server_address);
     }
+    // Mark connection to be drained
+    con->is_draining = 1;
     // Mark previous listen connection to be closed
     h->listen_connection->is_closing = 1;
     // Set the new listen connection
@@ -1784,7 +1786,6 @@ static json jsonrpc_machine_verify_send_cmio_response_state_transition_handler(c
 /// \param j JSON response object
 void jsonrpc_http_reply(mg_connection *con, http_handler_data *h, const json &j) {
     SLOG(trace) << h->server_address << " response is " << j.dump().data();
-    con->is_draining = 1;
     return mg_http_reply(con, 200, "Access-Control-Allow-Origin: *\r\nContent-Type: application/json\r\n", "%s",
         j.dump().data());
 }
@@ -1793,7 +1794,6 @@ void jsonrpc_http_reply(mg_connection *con, http_handler_data *h, const json &j)
 /// \param con Mongoose connection
 void jsonrpc_send_empty_reply(mg_connection *con, http_handler_data *h) {
     SLOG(trace) << h->server_address << " response is empty";
-    con->is_draining = 1;
     return mg_http_reply(con, 200, "Access-Control-Allow-Origin: *\r\nContent-Type: application/json\r\n", "");
 }
 
--- a/src/jsonrpc-remote-machine.cpp
+++ b/src/jsonrpc-remote-machine.cpp
@@ -33,7 +33,15 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
-#include <mongoose.h>
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#include "asio-config.h" // must be included before any ASIO header
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/signal_set.hpp>
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
+#include <boost/beast/version.hpp>
+#pragma GCC diagnostic pop
 
 #include "base64.h"
 #include "json-util.h"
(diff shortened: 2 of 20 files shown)


src/jsonrpc-virtual-machine.cpp
```diff
@@ -48,90 +57,153 @@ std::string jsonrpc_post_data(const std::string &method, const std::tuple<Ts...>
...
-        data->entity_body.clear();
-        data->status_code = "503";
-        data->reason_phrase = static_cast<char *>(ev_data);
-        data->done = true;
-    } else if (ev == MG_EV_CLOSE && !data->done) {
-        data->entity_body.clear();
-        data->status_code.clear();
-        data->reason_phrase = "connection closed";
-        data->done = true;
+static std::tuple<std::string, uint16_t, std::string> parse_url(const std::string &url) {
```
diegonehab, line 60: Use https://www.boost.org/doc/libs/1_85_0/libs/url/doc/html/url/urls.html?

edubart: I did this the first time, however Boost URL needs linking and my goal was to use header-only libraries, to make cross compilation easier. I think requiring to compile and link Boost libraries just for doing a small parsing is not justified. 

Boost URL used to have a header only mode before, but the author removed recently, saying that header-only libraries are bad because makes compilation slower.

src/jsonrpc-virtual-machine.cpp
```diff
@@ -48,90 +57,153 @@ std::string jsonrpc_post_data(const std::string &method, const std::tuple<Ts...>
...
+            } else { // Unexpected error
+                throw beast::system_error(ec);
+            }
+        }
+
+        // Set up HTTP response parser
+        beast::flat_buffer buffer;
+        http::response_parser<http::string_body> res_parser;
+        res_parser.eager(true);
+        res_parser.body_limit(16777216U); // can receive up to 16MB
```
diegonehab, line 158: Does this mean we will fail with a "read_memory" call for a larger chunk?

src/jsonrpc-virtual-machine.cpp
```diff
@@ -48,90 +57,153 @@ std::string jsonrpc_post_data(const std::string &method, const std::tuple<Ts...>
...
+                break;
+            } else if (ec == asio::error::interrupted) {
+                // Retry the operation during interrupts (SIGINT/SIGTERM),
+                // otherwise we may leave dead zombies processes during fork requests.
+            } else { // Unexpected error
+                throw beast::system_error(ec);
+            }
+        }
+
+        http::response<http::string_body> res = res_parser.get();
```
diegonehab, line 174: Is this a copy?

src/jsonrpc-virtual-machine.cpp
edubart: Yes. The same was true with Mongoose before, where we defined this:
```
# The biggest message is reset uarch with large_data: 2x 4MB + small data + overhad
MONGOOSE_DEFS+=-DMG_MAX_RECV_SIZE=12582912
```
So in our Mongoose implementation we could receive up to 12MB.

Boost Beast by default limits to 8MB, so this code duplicates the maximum, tests would fail with 8MB. If the message is bigger than that, the request will fail.

src/jsonrpc-virtual-machine.cpp
edubart: A shallow copy (the message buffer is not copied), but I will use `const&` .

edubart: Done.

src/jsonrpc-mgr.h
```diff
@@ -14,31 +14,34 @@
...
-#define JSONRPC_MG_MGR_H
+#ifndef JSONRPC_MGR_H
+#define JSONRPC_MGR_H
 
-#include <boost/container/static_vector.hpp>
-#include <mongoose.h>
 #include <string>
 
-namespace cartesi {
+#include "asio-config.h" // must be included before any ASIO header
```
diegonehab, line 22: This is producing warnings because of Boost's use of sprintf. This patch removes them.

```
diff --git a/src/jsonrpc-mgr.h b/src/jsonrpc-mgr.h
index dc7ee72..31de926 100644
--- a/src/jsonrpc-mgr.h
+++ b/src/jsonrpc-mgr.h
@@ -19,10 +19,13 @@
 
 #include <string>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include "asio-config.h" // must be included before any ASIO header
 #include <boost/asio/io_context.hpp>
 #include <boost/beast/core/tcp_stream.hpp>
 #include <boost/container/static_vector.hpp>
+#pragma GCC diagnostic pop
 
 namespace cartesi {
 
diff --git a/src/jsonrpc-remote-machine.cpp b/src/jsonrpc-remote-machine.cpp
index 85cd13a..bc697b0 100644
--- a/src/jsonrpc-remote-machine.cpp
+++ b/src/jsonrpc-remote-machine.cpp
@@ -33,12 +33,15 @@
 #include <sys/wait.h>
 #include <unistd.h>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include "asio-config.h" // must be included before any ASIO header
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/asio/signal_set.hpp>
 #include <boost/beast/core.hpp>
 #include <boost/beast/http.hpp>
 #include <boost/beast/version.hpp>
+#pragma GCC diagnostic pop
 
 #include "base64.h"
 #include "json-util.h"
diff --git a/src/jsonrpc-virtual-machine.cpp b/src/jsonrpc-virtual-machine.cpp
index 79731d2..ef0d35c 100644
--- a/src/jsonrpc-virtual-machine.cpp
+++ b/src/jsonrpc-virtual-machine.cpp
@@ -21,12 +21,15 @@
 #include <cstdint>
 #include <string>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #include "asio-config.h" // must be included before any ASIO header
 #include <boost/asio/connect.hpp>
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/beast/core.hpp>
 #include <boost/beast/http.hpp>
 #include <boost/beast/version.hpp>
+#pragma GCC diagnostic pop
 
 #include <boost/type_index.hpp>
 
```

src/jsonrpc-virtual-machine.cpp
diegonehab: For something small like this, this is fine. But we have a running disagreement on what is enough of a reason for you to reimplement functionality that exists elsewhere. Anyway, this seems to accept a string in the form `<host>:<port>/<path-etc>` with a mandatory `:<port>`. I don 't know why `/<path-etc>` is accepted. `<scheme>:` is not accepted and neither is the `//` before the `<host>`. If that's what was intended, just add a comment that this is not a url and we are done. I am sure the previous code was worse.

src/jsonrpc-virtual-machine.cpp
```diff
@@ -188,250 +258,261 @@ jsonrpc_mg_mgr::jsonrpc_mg_mgr(std::string remote_address) {
...
-semantic_version jsonrpc_virtual_machine::get_version(const jsonrpc_mg_mgr_ptr &mgr) {
+semantic_version jsonrpc_virtual_machine::get_version(const jsonrpc_mgr_ptr &mgr) {
     semantic_version result;
-    jsonrpc_request(mgr->get_mgr(), mgr->get_remote_address(), "get_version", std::tie(), result);
+    jsonrpc_request(mgr->get_stream(), mgr->get_remote_address(), "get_version", std::tie(), result);
     return result;
 }
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-parameter"
```
diegonehab, line 394: I'm almost running `git blame` on this to see who got tired of `(void)`. Was it me? 🤨

src/jsonrpc-mgr.h
edubart: Applied this diff.

src/jsonrpc-remote-machine.cpp
```diff
@@ -576,137 +722,113 @@ static json jsonrpc_get_version_handler(const json &j, mg_connection *con, http_
...
-        auto errno_copy = errno;
-        SLOG(error) << h->server_address << " fork failed (" << strerror(errno_copy) << ")";
-        mg_mgr_free(&h->child->event_manager);
-        delete h->child;
-        h->child = nullptr;
+    const int ret = fork();
+    if (ret == 0) { // child
+        // Notify to ASIO that we are the child
+        session->handler->ioc.notify_fork(asio::io_context::fork_child);
+        // Close all sessions there were initiated by the parent
```
diegonehab, line 774: there -> that

src/jsonrpc-remote-machine.cpp
```diff
@@ -576,137 +722,113 @@ static json jsonrpc_get_version_handler(const json &j, mg_connection *con, http_
...
-    // Mark connection to be drained
-    con->is_draining = 1;
-    // Mark previous listen connection to be closed
-    h->listen_connection->is_closing = 1;
-    // Set the new listen connection
-    h->server_address = new_server_address;
-    h->listen_connection = new_listen_connection;
+    const tcp::endpoint new_local_endpoint = address_to_endpoint(new_server_address);
+    if (new_local_endpoint.port() == 0) {
+        throw std::runtime_error{"rebind cannot be performed on port 0"};
```
diegonehab, line 807: Don't see why yes but neither why not.

diegonehab (commented):
This is looking good.

src/jsonrpc-virtual-machine.cpp
diegonehab: Should we limit the local API as well to 8M? Or do we allow them to differ in behavior?

edubart: I don't think we should limit the local API for reading/writing more than 16MB, this is something I personally do for local machines (for testing and hacking purposes). The remote machine is mostly used by the node, and I don't think the node will ever need to read more than 16MB, so I personally don't mind having them different for this specific case. If someone complains we could raise the limit in the future.

diegonehab: I am worried about the send_cmio_response call. Who knows when somebody will want to dehash a large chunk of data? We should keep this in mind and advertise the limitation. Not sure either if the node will use jsonrpc or not.

src/jsonrpc-remote-machine.cpp
edubart: Because the `rebind` request doesn't return the new endpoint to connect to, the client would not know the new address to connect to, so it wouldn't make sense to allow it to perform.

src/jsonrpc-virtual-machine.cpp
edubart: Good point, I was not happy with the URL parsing either. I removed the extra URL related code, because it's not really a URL, the old code plus Boost documentation influenced me to parse as URL, now it is simpler.

diegonehab: The parsing now is a bit weird. It accepts `<host>:<port>` or just `<host>` if `<host>` is a starts with a digit. In which case it will use `<host>` as the host, and the numeric prefix in `<host>` as the port. Right? 😛

src/jsonrpc-virtual-machine.cpp
diegonehab: What kind of error message does the client get when trying to, say, write_memory(20MB)?

src/jsonrpc-virtual-machine.cpp
edubart: Indeed, I've improved the parsing to check better malformed addresses.

diegonehab (approved):

mpernambuco (approved):

PR #241, opened 2024-05-17
Publish patch that adds generated files (uarch.bin etc) to source tree

diegonehab:
As we move towards providing packages for cartesi-machine-emulator for multiple platforms, we should make the job of package maintainers as easy as possible. One difficulty in building our emulator is its dependency on uarch.bin, which itself depends on a special toolchain we distribute inside a Docker image. 

Instead of forcing package maintainers to deal with these issues, this PR introduces a new target, `create-generated-files-patch` that produces a patch `add-generated-files.diff` that, when applied to the source tree, introduces the files needed to build the emulator.

It also includes changes to CI that cause `add-generated-files.diff` file to be published as an artifact on tagged releases.


mpolitzer:
Released artifacts now have mixed version / no version suffixes, can we clean this up here or on a new issue?
My vote would be to remove them from the files and leave it only on the URL when downloading.

diegonehab:
I think we can add the version names to the files. Or remove them. But the patch will put files without versions into the source tree. It could be another issue as well. @vfusco @edubart

edubart:
Like @mpolitzer, I also vote for removing versioning of file names. Versioned filenames are becoming a pain in some scripts and also during packaging because we need to propagate versions through CI, scripts and makefiles.

.github/workflows/build.yml
```diff
@@ -102,6 +102,8 @@ jobs:
             uarch-ram.bin
             uarch-pristine-ram.c
             uarch-pristine-hash.c
```
edubart, line 104: Should we remove these `.c` and `.h` files now? Since the diff fulfills the purpose? Somewhat redundant.

.github/workflows/build.yml
```diff
@@ -102,6 +102,8 @@ jobs:
...
+            machine-c-version.h
+            add-generated-files.diff
```
edubart, line 106: Why not `.patch` instead of `.diff` ? Usually the name extensions is `.patch` when packaging in most Linux distros.

diegonehab:
Closed by mistake.

diegonehab:
Opening again.

PR #243, opened 2024-05-21
Publish patch that adds generated files (uarch-pristine-hash.c etc) to source tree

diegonehab:
As we move towards providing packages for cartesi-machine-emulator for multiple platforms, we should make the job of package maintainers as easy as possible. One difficulty in building our emulator is its dependency on uarch.bin, which itself depends on a special toolchain we distribute inside a Docker image.

Instead of forcing package maintainers to deal with these issues, this PR introduces a new target, create-generated-files-patch that produces a patch add-generated-files.diff that, when applied to the source tree, introduces the files needed to build the emulator.

It also includes changes to CI that cause add-generated-files.diff file to be published as an artifact on tagged releases.


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Makefile target `create-generated-files-patch` to generate `add-generated-files.diff` patch
+- CI now creates and uploads `add-generated-files.diff` as artifact on tagged releases
+- CI uploads `machine-c-api.h` as artifact on tagged releases
```
mpolitzer, line 12: Not anymore right?

diegonehab: Updated.

mpolitzer (approved):

.github/workflows/build.yml
```diff
@@ -710,16 +711,23 @@ jobs:
...
         uses: softprops/action-gh-release@v1
         if: ${{ startsWith(github.ref, 'refs/tags/v') }}
         with:
           draft: true
           files: |
             artifacts/cartesi-machine-*.deb
             artifacts/uarch-ram.bin
-            artifacts/uarch-pristine-ram.c
-            artifacts/uarch-pristine-hash.c
+            add-generated-files.diff
```
edubart, line 730: I think it is way common to name `.patch` instead of `.diff`, here is an example for aports (Alpine package tree):

```
bart@archawk:~/repos/aports (ltrace-riscv64 %) $ find -name '*.patch' | wc
   5171    5171  237831
bart@archawk:~/repos/aports (ltrace-riscv64 %) $ find -name '*.diff' | wc
      6       6     295
```

So I suggest following the standard.

PR #246, opened 2024-06-27
fix: always read valid TLB entries while computing hash with uarch

edubart:


mpernambuco (dismissed):

edubart:
Rebased and added guards to check if TLB is not in a corrupt state, so next time this happens we get an exception instead of a crash.

mpernambuco (approved):

mpolitzer (approved):

vfusco (approved):

PR #247, opened 2024-06-27
fix: ensure make clean works with no submodules

mpolitzer:
without this patch `make clean` doesn't work for source tarballs.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -130,7 +130,7 @@ dependencies.sha256:
 	@shasum -a 256 $(BUILDDIR)/images/* > $@
 
 clean:
-	@$(MAKE) -C ../third-party/riscv-arch-test/riscv-target/cartesi clean
+	@$(MAKE) -C ../third-party/riscv-arch-test/riscv-target/cartesi clean || true
 	@$(MAKE) -C misc clean
 	@rm -f *.profdata *.profraw *.gcda *.gcov
 	@rm -rf $(BUILDDIR)


tests/Makefile
```diff
@@ -130,7 +130,7 @@ dependencies.sha256:
 	@shasum -a 256 $(BUILDDIR)/images/* > $@
 
 clean:
-	@$(MAKE) -C ../third-party/riscv-arch-test/riscv-target/cartesi clean
+	@$(MAKE) -C ../third-party/riscv-arch-test/riscv-target/cartesi clean || true
```
edubart, line 133: I think we should check if directory exists, instead of using `||`.

tests/Makefile
```diff
@@ -130,7 +130,7 @@ dependencies.sha256:
 	@shasum -a 256 $(BUILDDIR)/images/* > $@
 
 clean:
```
edubart, line 132: What about the [fix-removal-of-generated-files](https://github.com/cartesi/machine-emulator/pull/249/commits/fb679a6cda31955301f58f988b5f984b97f71eff#diff-6bb23a60a80d75ac81e61f9ae0518579761646eac951cca846625688fbfd0380) you did in the other PR? Can we also have the changes you needed from that PR in this one?

mpolitzer: Yes, I'll update it.

edubart (approved):

tests/Makefile
mpolitzer: I removed [fix-removal-of-generated-files](https://github.com/cartesi/machine-emulator/pull/249/commits/fb679a6cda31955301f58f988b5f984b97f71eff#diff-6bb23a60a80d75ac81e61f9ae0518579761646eac951cca846625688fbfd0380) since it is not a bug and only makes sense in the debian package context. I'll move it there instead.

vfusco (approved):

edubart (approved):

PR #249, opened 2024-07-01
feat: add debian source package files

mpolitzer:
# workflow:
```
# install debian packaging tools
apt install -y --no-install-recommends build-essential dh-make devscripts 

# update the version on the changelog
dch -v `make -s version` "Release `make -s version`-1"

# recreate the debian./ folder and the  ".orig" tarball in ..
dh_make -y -l --createorig -c lgpl3 -p machine-emulator_`make -s version`

# or just the tarball
tar czf ../machine-emulator_`make -s version`.orig.tar.gz --exclude=.git .

# build the dsc
dpkg-source -b .

# build the binary package
dpkg-buildpackage -us -uc

# cleanup
dpkg-buildpackage -rfakeroot -Tclean
```


debian/control
```diff
@@ -0,0 +1,43 @@
...
+Build-Depends:
+ debhelper-compat (= 13),
+ libboost1.81-dev,
+ liblua5.4-dev,
+ libslirp-dev,
+ lua-dkjson,
+ lua-lpeg,
+ lua-posix-dev,
+ lua-sec-dev,
+ lua-socket-dev,
```
edubart, line 15: `lua-posix`, `lua-sec`, `lua-socket` are only needed for testing could remove if we are not performing tests.
`lua-lpeg` can be removed
`lua-dkjson` we are not using anymore for now

debian/control
```diff
@@ -0,0 +1,43 @@
...
+#Vcs-Git: https://salsa.debian.org/debian/machine-emulator.git
+
+Package: machine-emulator-dev
+Section: libdevel
+Architecture: any
+Multi-Arch: same
+Depends:
+ libboost1.81-dev,
+ liblua5.4-dev,
+ libslirp-dev,
```
edubart, line 29: The `lua-dev` and `slirp-dev` are only needed during build time, so could move to `Build-Depends`.

debian/patches/series
```diff
@@ -0,0 +1,3 @@
+fix-make-clean
+add-generated-files
+fix-removal-of-generated-files
```
edubart, line 3: Should probably make the changes needed for `fix-removal-of-generated-files` and `fix-make-clean` into the emulator source tree.

edubart:
Could the workflow shared on this PR comment be a workflow in CI?

debian/control
mpolitzer: I took them from the Dockerfile, I'll add an issue to clean it up as well.

mpolitzer:
> Could the workflow shared on this PR comment be a workflow in CI?

Yes, I plan on replacing the binary version with this one on CI.

I just wanted to confirm the direction since the debian source packaging is so intrusive. Thus the draft.

edubart:
We are not going to do packaging inside this repository anymore, it will be on `linux-packages` repo, so I am closing the PR.

I am already providing packages for `cartesi-machine-emulator` there.

PR #250, opened 2024-07-01
chore: compile in release mode with debug info by default

edubart:
This PR switch the default compilation from release with asserts enabled for everything, to release with debug information with asserts disabled only for `interpret.cpp`. This makes sure people will compile an efficient emulator with just `make` when packaging by default, not requiring the use of `make relwithdebinfo=yes` or `make release=yes` anymore when packaging.

The default make is chosen to contain debug information because some distros may want to strip and move the debug information into a second debug package (for example ArchLinux does this).

Change were mades in CI to always test with debug information, except for tagged releases.

**IMPORTANT:** After this PR, if you want to compile with asserts for `interpret.cpp`, please compile with `make debug=yes`, this is something only emulator developers care.

## PR main changes
- Compile `interpret.cpp` with `-DNDEBUG` for release/relwithdebinfo modes
- Compile with `relwithdebinfo=yes` by default, but use `debug=yes` for CI tests
- Always compile `sha3.c` (third party source) with `-O3` to speed up ci
- Use `-O1` optimizations on sanitize action to make CI faster
- Add `libcartesi.so` symlinks in `src`, to make linkage on external programs dependent on in-development emulator straightforward

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,10 +113,21 @@ DOCKER_PLATFORM=--platform $(BUILD_PLATFORM)
 endif
 
 # Code instrumentation
+debug?=no
+relwithdebinfo?=no
 release?=no
 sanitize?=no
 coverage?=no
+
+# If not build type is chosen, set the default to release with debug information,
+# so the emulator is packaged correctly by default.
+ifeq (,$(filter yes,$(relwithdebinfo) $(release) $(debug) $(sanitize)))
+relwithdebinfo=yes
+endif
+
 export sanitize
+export debug
+export relwithdebinfo
 export release
 export coverage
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -113,10 +113,14 @@ DOCKER_PLATFORM=--platform $(BUILD_PLATFORM)
 endif
 
 # Code instrumentation
+debug?=no
+relwithdebinfo?=no
```
mpolitzer, line 117: Please use `relwithdbginfo` or `relwithdebuginfo`.
`deb` is sometimes used for`debian` and may cause confusion.

edubart: The compile mode names are borrowed from CMake, where it have `Release`, `Debug,` `RelWithDebInfo` and `MinSizRel` https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html . Although we don't have to follow it, I thought using the name that is already being used would be less surprise. Also this name has already being used before this commit.

mpolitzer: ok, following some naming convention seems better than trying to avoid this confusion.

mpolitzer:
Tested the different builds and the flags look good.
A suggesting is to add those options to the `make help` target. Something like:
```
help:
	@echo 'Main targets:'
	@echo '* all - Build the src/ and uarch/ code. To build from a clean clone, run: make submodules all'
	@echo '        Additionally, to specify the build type, pass yes to one of:'
	@echo '        - sanitize, debug, relwithdebinfo, *release*, coverage. e.g: make sanitize=yes
```

Another point, a bit unrelated to this PR is removing the Emscripten documentation from the README.md. A brief comment pointing to the wiki should suffice. Should we do the same for the mingw toolchain?

mpolitzer (dismissed):

edubart:
> Another point, a bit unrelated to this PR is removing the Emscripten documentation from the README.md. A brief comment pointing to the wiki should suffice. Should we do the same for the mingw toolchain?

I created issue https://github.com/cartesi/machine-emulator/issues/261 for me to improve README/help later (to not delay release).

mpolitzer (approved):

PR #253, opened 2024-07-02
refactor!: design new simplified C API

edubart:
This PR is very **WIP**, it will be a long road before this is finished and polished. I created the PR before the time so I can document here a changelog of what is being changed while I change them. Also to request for comments on the changes.

You can look around the C API files to see how the new simplified API feels. I have also update each function documentation along the way.

## Changes

Added:
- Added `CM_MCYCLE_MAX` macro and set to `UINT64_MAX`
- Added `CM_ACCESS_LOG_TYPE` enum with bitwise flags for access log type
- Added `cm_get_last_error_message`/`cm_jsonrpc_get_last_error_message`, returning the error message for the most recent failed call.
- Added `cm_get_default_runtime_config`/`cm_jsonrpc_get_default_runtime_config`, for now it will return just `{}`, but we might in the future return a value based on current environment variables, like the Cartesi Node does. People also asked for this function.
- Added views CSRs for HTIF fromhost/tohost for dev/data/cmd (eg. `CM_CSR_HTIF_TOHOST_DEV`), the idea is to write on fields of these CSRs through `cm_write_csr`/`cm_read_csr`.
- Added views CSRs for IFLAGS PRV/X/Y/H (eg. `CM_CSR_IFLAGS_H`), the idea is to write on fields of these CSRs through `cm_write_csr`/`cm_read_csr`.

Removed:
- Remove all structs, complex structs are passed as JSON strings now (configs, access logs and proofs).
- Remove all `cm_read_*`/`cm_write_*` of CSRs, except for `mcycle`/`uarch_cycle`/`uarch_halt_flag`.
- Remove all `cm_delete_*` functions, now any function that returns a string guarantees to maintain the validity of the string pointer until next API call or machine destruction.
- Removed `cm_new_default_machine_config`
- Removed `cm_reset_iflags_*`/`cm_set_iflags_*`/`cm_packed_iflags`, `iflags` can be read/written using CSR views
`CM_CSR_IFLAGS_PRV`, `CM_CSR_IFLAGS_X`, `CM_CSR_IFLAGS_Y`, `CM_CSR_IFLAGS_H`.
- Removed the `machine-c-defines.h` header, its contents was moved to `machine-c-api.h`.

Renames:
- Renamed `*_uarch_reset*` -> `*_reset_uarch*`
- Renamed `*_uarch_step*` -> `*_step_uarch*`
- Renamed `cm_machine_run` -> `cm_run`
- Renamed `cm_create_machine` -> `cm_create`
- Renamed `cm_load_machine` -> `cm_load`,
- Renamed `cm_machine_run_uarch` -> `cm_run_uarch`
- Renamed `CM_PROC_CSR` -> `CM_CSR`
- Renamed `cm_create_jsonrpc_machine` -> `cm_jsonrpc_create_machine`
- Renamed `cm_load_jsonrpc_machine` -> `cm_jsonrpc_load_machine`
- Renamed `cm_get_jsonrpc_machine` -> `cm_jsonrpc_get_machine`
- Renamed `cm_create_jsonrpc_mgr` -> `cm_jsonrpc_create_mgr`
- Renamed `cm_delete_jsonrpc_mgr` -> `cm_jsonrpc_destroy_mgr`

Changed:
- All all machine runtime config arguments were removed from verify functions, because seems like they were not being used.
- All `err_msg` arguments were removed, the latest error message can be retrieved from `cm_get_last_error_message()`.
- Changed `cm_replace_memory_range`, now it takes multiple arguments instead of a struct argument.
- Reorganized the C API intro different categories in the header, making easy for the reader to navigate through the API.
- Replaced all definitions with enums, because they are more friendly for binding generators.
- Polished the API documentation.

## Changes rationale
First have in mind the goal of the changes was to simplify usage of the C API in other languages, to minimize the binding work and make easier to create binding generators.

- Overall the C API decreased from 152 functions to 59 functions.
- The rationale of removing `_machine` from some methods is that most API are related to the machine yet they don't contain the machine word, so here we make this the standard.
- All methods related to jsonrpc were forced to always begin with `cm_jsonrpc_`, this makes it easier for binding generators to categorize and filter jsonrpc related methods.
- All structs were removed, because binding structs in some languages can be quite a work, also because we will now use JSON strings for complex objects (configs, access logs and proofs).
- All methods that return a C string (for JSON example) will guarantee validity of the string until the same method is called again, saving the need of any deallocating string functions. Most bindings will parse the JSON immediately and duplicate its contents into a special JSON format the language cares about anyway, it's not like people will retain the string pointer stored somewhere, even so the documentation explicitly states its lifetime. This also removes the burden of developers to do C memory management.
- The only methods required to call to free memory are `cm_destroy` and `cm_jsonrpc_destroy_mgr`.


src/machine-c-api.h
```diff
@@ -451,1508 +159,435 @@ typedef struct { // NOLINT(modernize-use-using)
...
-/// \param val New register value.
-/// \param err_msg Receives the error message if function execution fails
-/// or NULL in case of successful function execution. In case of failure error_msg
-/// must be deleted by the function caller using cm_delete_cstring.
-/// err_msg can be NULL, meaning the error message won't be received.
+/// \brief Returns count of memory ranges in the machine.
+/// \param count Receives the memory range count.
 /// \returns 0 for success, non zero code for error
-CM_API int cm_write_misa(cm_machine *m, uint64_t val, char **err_msg);
+CM_API int cm_get_memory_range_count(cm_machine *m, int *count);
```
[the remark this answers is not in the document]
edubart: Cleaver trick.

But one goal here is to not ask users who is using the API or creating a binding to perform allocations on their own, some might not know how to use `malloc/free` well, in fact some people using the C API of the emulator confused before who has the ownership of these struct pointers, either they forget to free thinking the ownership moved, or caused double free.

If we do the suggested cleaver way, it would also be the only struct needed, which would increase the logic of the binding generators just because of one struct, we might as well get rid of it. Ideally I was thinking of an API like OpenGL, who gets away without any struct. The proposed memory range API I did has two functions but I think it is more fail proof of users mistakes and more easy on the binding generator work despite adding a new function.  Also people binding the emulator C API complained that the binding work of structs are boilerplate.

src/machine-c-api.h
```diff
@@ -82,1894 +83,544 @@ typedef enum { // NOLINT(modernize-use-using)
...
+/// \brief Replaces a memory range.
+/// \param m Pointer to a valid machine instance.
+/// \param start Memory start physical address.
+/// \param length Memory length.
+/// \param shared If true changes by the machine will be shared to the image file.
+/// \param image_filename Image file name to load into the range.
+/// \returns 0 for success, non zero code for error.
+/// \details The machine must contain an existing memory range matching the start and length
+/// specified in new range.
+CM_API int cm_replace_memory_range(cm_machine *m, uint64_t start, uint64_t length, bool shared,
```
diegonehab, line 222: This function is often used to clear a memory range, so it makes sens t o allow for `NULL` `image_filename` for that purpose.

src/machine-c-api.h
```diff
@@ -82,1894 +83,544 @@ typedef enum { // NOLINT(modernize-use-using)
...
 /// \brief Runs the machine for one micro cycle logging all accesses to the state.
-/// \param m Pointer to valid machine instance
-/// \param log_type Type of access log to generate.
+/// \param m Pointer to a valid machine instance.
+/// \param log_type Type of access log to generate, must be bitwise OR of CM_ACCESS_LOG_TYPE enum.
+/// \param one_based Use 1-based indices when reporting errors.
+/// \param access_log Receives the state access log as a JSON string,
+/// remains valid until this same function is called again for the same machine.
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_log_uarch_step(cm_machine *m, int log_type, bool one_based, const char **access_log);
```
diegonehab, line 268: Should be `cm_log_step_uarch`, right? Because if there was a function to step the uarch without returning a log, it would have been `cm_step_uarch` instead `cm_uarch_step`. Other functions would change accordingly.

src/machine-c-api.h
```diff
@@ -82,1894 +83,544 @@ typedef enum { // NOLINT(modernize-use-using)
...
+/// \param m Pointer to a valid machine instance.
+/// \param address Physical address (aligned to 64-bit boundary).
+/// \param value Receives the value of the word.
+/// \returns 0 for success, non zero code for error.
+/// \warning The current implementation of this function is very slow!
+CM_API int cm_read_word(const cm_machine *m, uint64_t address, uint64_t *value);
+
+/// \brief Write the value of a word in the machine state.
+/// \param m Pointer to a valid machine instance.
+/// \param address Physical address (aligned to 64-bit boundary).
```
diegonehab, line 383: This function was not there for an obvious reason: the read function can read any word in the state, include those mapped to CSRs and devices. Will this function be able to do the same? See the problem?

src/machine-c-api.h
```diff
@@ -82,1894 +83,544 @@ typedef enum { // NOLINT(modernize-use-using)
...
+/// \param m Pointer to a valid machine instance.
 /// \param val Receives the value of the register
-/// \param err_msg Receives the error message if function execution fails
-/// or NULL in case of successful function execution. In case of failure error_msg
-/// must be deleted by the function caller using cm_delete_cstring.
-/// err_msg can be NULL, meaning the error message won't be received.
-/// \returns 0 for success, non zero code for error
-CM_API int cm_read_pc(const cm_machine *m, uint64_t *val, char **err_msg);
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_read_pc(const cm_machine *m, uint64_t *val);
```
diegonehab, line 435: This is not very common to use, right? Except for the debugger perhaps? Shouldn't we remove it? You can always use read_csr, or even write your own convenience function for it. Same for `write_pc`.

src/jsonrpc-machine-c-api.h
```diff
@@ -17,240 +17,205 @@
...
+/// \param mgr Pointer to a valid jsonrpc connection manager.
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_jsonrpc_shutdown(const cm_jsonrpc_mgr *mgr);
+
+/// \brief Gets the semantic version of remote server machine
+/// \param mgr Pointer to a valid jsonrpc connection manager.
+/// \param major Receives semantic major version.
+/// \param minor Receives semantic minor version.
+/// \param patch Receives semantic patch version.
+/// \param pre_release Receives pre release string,
```
diegonehab, line 70: I don't like this behavior or "something being valid until a function overwrites it". It is bug prone.
It's the kind of thing that the C standard has been fighting for decades to get rid of (see _r variants).

I think that the users having to check the manual for the lifetime of string that functions return is worse than telling they always get their own copy that they should free (or not, who cares). I prefer the user to always do the same thing: got a string? free it. It's yours.

There really is no concrete performance issue here that I can see.

The read-memory function is the only one that could potentially be called repeatedly, and that already receives a buffer to write in place, right?

I understand your urge to avoid the need for users to call "free" on the other side, but do we really need to be so paternalistic?

src/jsonrpc-machine-c-api.h
```diff
@@ -17,240 +17,205 @@
...
+/// remains valid until this same function is called again for the same machine.
+/// \param build Receives build string,
+/// remains valid until this same function is called again for the same machine.
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_jsonrpc_get_semantic_version(const cm_jsonrpc_mgr *mgr, uint32_t *major, uint32_t *minor, uint32_t *patch,
```
diegonehab, line 75: JSON as well?

src/machine-c-api.h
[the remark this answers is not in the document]
diegonehab: Just return a JSON with all the fields there.

src/machine-c-api.h
edubart: Makes sense, will add this to its documentation.

src/machine-c-api.h
edubart: I imagined, can be tricky to implement. Will remove it then.

src/machine-c-api.h
edubart: Ok, will remove.
So we implement only for `mcycle`/`uarch_cycle`/`uarch_halt_flag` ? Or should also remove the function for `uarch_halt_flag` ?

src/jsonrpc-machine-c-api.h
edubart: Ok, will return a JSON instead.

diegonehab (changes requested):
I understand the urge to change break iflags apart at the moment, but I have a different proposal that I'd like not to mix with this.

At the moment, iflags has `PRV` and the `X`, `Y`, and `H` flags.

`PRV` really is something internal that the host should never mess with.

`X`, `Y`, and `H`, on the other hand, are things the host needs to look at and (in the case of `Y`, change).

There is never a case in which more than one of these flags is set.

They are always set via HTIF from the inside.

In the case of `X` and `Y`, the host will also need to look into  `htif.tohost`.

The obvious place for these flags to live in is in `htif.tohost` itself.

With some reorganization there, we can make this happen.

We would rename `iflags` to `iprv` and make it a full CSR without any packing/unpacking or dedicated read/write methods.

The interpreter loop would check for the equivalent of `X`, `Y` and `H` in `htif.tohost`. While we are at it, let's rename `X` to `YA` (or `YO`) and `Y` to `YM` (or `YI`). 

After a return from run, the host would already have the reason.
The host would read `htif.tohost` and decide what to do.
There already are many `WARL` CSRs that prevent certain bits from being changed. 
`htif.tohost` would be one of these. If `H` is set, it would remain set forever. I think we can even use a write to `htif.fromhost` to clear `Y`, saving the need to modify `htif.tohost`. 

`X` is cleared automatically. But we could we can remove `X`, right? There is no need for it anyway, now that we return the status in the call to run? 

Anyway, the scope of these changes are way outside of the API simplification for binds. I would like to avoid mixing these.

src/machine-c-api.h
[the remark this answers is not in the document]
edubart: I was trying to avoid returning a JSON for this specific case, otherwise people will really need to include a JSON parser in case they want to iterate over a machine PMAs. The same problem happens for jsonrpc semantic version return.

The proposed API changes a C user could use most of the API without needing to include a JSON parser library.

Thinking about people using the API with just C, asking them to depend on a JSON parser is kinda annoying in my view, but well, maybe they will not need to iterate over PMAs anyway.

In my view JSON strings only make sense for configs, access logs and proofs, not more than that.

src/machine-c-api.h
diegonehab: I think Dave will the halt flag frequently, might as well keep it.

diegonehab: See comments on `htif_tohost` and `htif_fromhost` and `iflags_*.*`. I think we should keep them as is for now. But do a radical change later on, where we'd get rid of all `iflags` functions, and keep only `read_htif_tohost` and `write_htif_fromhost`.

src/jsonrpc-machine-c-api.h
edubart: > I understand your urge to avoid the need for users to call "free" on the other side, but do we really need to be so paternalistic?

It makes the C API easier to use in my view, because it becomes less boilerplate (less memory management calls imposed for the user to do). It's more work for the user to free something immediately after a function call, and I think it is also bug prone either way in case users forget to free, or try to be clever deferring its deallocation sometime in the future. Users have to read the documentation either way. I have seen people struggle with C memory management, one of the goals here was to minimize the burden to do this.

But what are your suggestions here? Ask for the user to free every returned string in the API? What about error messages? Would they be an exception so we can have `cm_get_last_error_message()` ? Or you think we should go all the way and keep the same behavior of every function having an `err_msg` which needs to be freed?

src/machine-c-api.h
edubart: > ut do a radical change later on, where we'd get rid of all iflags functions
Sure, I will avoid changing how `iflags` works.

edubart:
Addressed some comments:
- Avoid splitting IFLAGS, introduce CSR views of its fields instead, to avoid creating unpack/pack methods.
- Removed addition of `cm_write_word`.
- Improved documentation of `cm_replace_memory_range`
- Renamed `*_uarch_step*` -> `*_step_uarch*`
- Removed read/write for CSRs `pc` and `uarch_pc`.

src/machine-c-api.h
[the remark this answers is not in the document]
mpolitzer: My worry is that the API won't be ergonomic for "lower" level languages, just the high level ones.
But ok, you comment makes sense since you managed to remove the other structs.

src/jsonrpc-machine-c-api.h
mpernambuco: My vote in favor of Implementing `cm_get_last_error_message()`. It will alleviate 90% of the pain that users face when using the C API.

src/machine-c-api.h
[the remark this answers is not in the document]
edubart: > My worry is that the API won't be ergonomic for "lower" level languages
Agreed, but calling these 2 functions still easy to use in low level languages. However returning JSON it's less easy because you have to pull a JSON parser.

src/jsonrpc-machine-c-api.h
mpernambuco: If the API docs states, 
"Strings are returned in pointers that remain valid and hold the same value until the next time the same API function is called"
, then the user only needs to read it once. 
Not having to free memory will save them a lot of time. 
This approach is obviously easier and not more error  prone than expecting the user to free memory."

diegonehab: That's for sure, @mpernambuco . `cm_get_last_error_message` simplifies the use of *every* other API function, and we are keeping it.

What we are discussing here is basically the management of strings returned by the API. Can the API return a string that points to some internal memory buffer kept by the binding or the machine class, and that will be overwritten or potentially deleted by the binding or machine class at some sort-of-unkown time in the future? Can some other function return a string that points to an internal memory buffer that will remain valid until the machine object is destroyed, and that the user can't modify or that same function might return the (now incorrect) modified string back?

I think that `cm_get_last_error_message` retains its value even if the string it returns is a copy and the user is forced to free it afterwards. This is because it allows error handling to be separated from the place where a function that failed was called. I.e., it can be done only once and called many times.

edubart: Another idea, I could ask the user to pass a buffer size and length to fill with the JSON, and document a minimum size the buffer must have. Would be this way for all functions returning a JSON, what you think? This way there is no allocation ownership being moved from our API to the user.

diegonehab: I think we could make `cm_get_last_error_message` an exception, or we could make it return a copy. What I don't want is to open the room for users to think they have a string that isn't modified and then have the string modified behind their back, and I don't want them to constantly have to check if the string can be modified or not or if they have to free or not depending on the function they called.

src/machine-c-api.h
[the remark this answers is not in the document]
diegonehab: Either we have the structs, or we don't. If we don't have the structs, you need a JSON parser. I don't see the distinction here. If we are going to do it, let's go all the way. Or are you saying that iterating over the memory ranges is something that is more common than getting the default config or getting a proof or getting a step log or some such?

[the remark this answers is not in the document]
diegonehab: I am assuming every bind will be to a language that has JSON support. The only language that will suffer is C, right? Do we have examples other than C? Is anybody really going to use C?

src/jsonrpc-machine-c-api.h
diegonehab: > Another idea, I could ask the user to pass a buffer pointer and its length to fill with the JSON, and document a minimum size the buffer must have. Would be this way for all functions returning a JSON, what you think? This way there is no allocation ownership being moved from our API to the user responsibility.

It's difficult to estimate the size needed without actually performing the action. The log will be a problem, for example.

diegonehab: > If the API docs states, "Strings are returned in pointers that remain valid and hold the same value until the next time the same API function is called" , then the user only needs to read it once. Not having to free memory will save them a lot of time. This approach is obviously easier and not more error prone than expecting the user to free memory."

So we will have an internal string buffer for each API function that returns a string, in thread local storage? We could do that.

diegonehab:
> Addressed some comments:
> 
>     * Avoid splitting IFLAGS, introduce CSR views of its fields instead, to avoid creating unpack/pack methods.
> 
>     * Removed addition of `cm_write_word`.
> 
>     * Improved documentation of `cm_replace_memory_range`
> 
>     * Renamed `*_uarch_step*` -> `*_step_uarch*`
> 
>     * Removed read/write for CSRs `pc` and `uarch_pc`.

I think we should keep `reset_iflags_Y` at least. And we might want to keep `read_iflags_X`, 'read_iflags_Y`, and `read_iflags_H` just because everybody uses them.

src/jsonrpc-machine-c-api.h
edubart: > So we will have an internal string buffer for each API function that returns a string, in thread local storage? We could do that.

That was the original idea for most functions in this PR if that was not clear. Every function returning a JSON has a per function storage, which is retained until the same function is called again, I will update the documentation and make that more clear.

src/machine-c-api.h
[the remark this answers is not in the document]
mpolitzer: > I am assuming every bind will be to a language that has JSON support. The only language that will suffer is C, right? Do we have examples other than C? Is anybody really going to use C?

Even C has it, but that is not the point.
Go and Rust handle C structs great, Dynamic languages maybe not so well (However Node could bind the C++ API directly), Lua is already contemplated, which leaves out Python.

I believe that manual memory management is more of an issue since none of the above has it. At least from what I remember on the subject.

This opinion may be a little late on this discussion, but my worry is that we are optimizing for the wrong thing.

diegonehab:
One thing to consider. Are we doing to offer a “raw” version of these JSON functions in the binds, ones that return/receive JSON and other that receive/return parsed objects (like lua tables) or not?

edubart:
> One thing to consider. Are we doing to offer a “raw” version of these JSON functions in the binds, ones that return/receive JSON and other that receive/return parsed objects (like lua tables) or not?

I think I could try making a generic function to translate  JSON <-> Lua table, so we don't use raw JSON strings on Lua side. In the worst case if that turns out to be a bit of work or having flaws (such as JSON special values like its NULL), we could have a raw JSON string and on Lua side and use `lua-dkjson`. For now my concerns is the C API, later I will think more about Lua goodies.

edubart:
I updated the PR adding back some `iflags` methods and trying to clarify more how string lifetime for JSON will work (thread local storage per function).

src/jsonrpc-machine-c-api.h
edubart: Read again the PR, I tried to clarify the lifetimes. It's like Perna said, remains valid until next call using thread local storage.

diegonehab: What happens when one thread deletes the machine?

src/machine-c-api.h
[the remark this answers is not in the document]
edubart: > Either we have the structs, or we don't. If we don't have the structs, you need a JSON parser. I don't see the distinction here. If we are going to do it, let's go all the way. Or are you saying that iterating over the memory ranges is something that is more common than getting the default config or getting a proof or getting a step log or some such?

I think it is, if I use the C API in C program, I might not care about getting initial configs, or proofs, or access log of a stored machine. But I might care checking what memory ranges the loaded machine has.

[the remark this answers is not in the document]
diegonehab: This is something you will do once per machine you load. It’s not at issue regarding performance. Nobody will use the C API directly.

src/jsonrpc-machine-c-api.h
edubart: > What happens when one thread deletes the machine?

The strings will still remain valid and allocated, until next call for the same function  despite machine descrution. After its first call, the string will be only be ever deallocated on thread termination via a global desctructor.

src/machine-c-api.h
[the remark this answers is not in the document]
edubart: Ok, I will change this to be also a JSON.

src/jsonrpc-machine-c-api.h
diegonehab: So if you have two machines on a thread, the same function call on different machines will overwrite each other?

diegonehab: Using C++ `thread_local`?

edubart: > So if you have two machines on a thread, the same function call on different machines will overwrite each other?

Correct, for example, in the same thread calling `cm_get_initial_config()` twice, the second call will overwrite contents returned by the first call. If a user wants to retain the first call contents after the second call, he will need to duplicate the first call string himself.

But as I mentioned, these are JSON strings, most likely these strings will be parsed right-away into to some internal language specific JSON data structure that will be able to retain its values. Is not like people will keep holding and working with JSON C strings around, it's a pain to work with JSON as C string, languages usually duplicate them anyway to some other internal JSON object.

> Using C++ thread_local?
Correct.

src/machine-c-api.h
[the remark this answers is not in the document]
edubart: Removed the two methods and introduced `cm_get_memory_ranges` returning a JSON string in latest commit.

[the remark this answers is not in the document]
diegonehab: Sure!

src/jsonrpc-machine-c-api.h
diegonehab: So this will be the behavior of the machine class in C++ as well, right? Return const ref to a thread_local string.

edubart: > So this will be the behavior of the machine class in C++ as well, right? Return const ref to a thread_local string.

Could be, but it's not necessary. I could do this outside C++ API, just in the C API binding file.

diegonehab: When we say "overwritten" we mean "invalidated", right. As in, the pointer would now be dangling. I don't think we can avoid that. In that case, we might as well move this thread_local complexity to the C++ class itself, to simplify the C API implementation (which, BTW, could later use some love...)

src/machine-c-api.h
```diff
@@ -82,1894 +83,528 @@ typedef enum { // NOLINT(modernize-use-using)
...
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_log_reset_uarch(cm_machine *m, int log_type, bool one_based, const char **access_log);
+
+/// \brief Sends cmio response.
+/// \param m Pointer to a valid machine instance.
+/// \param reason Reason for sending the response.
+/// \param data Response data to send
+/// \param length Length of response data.
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_send_cmio_response(cm_machine *m, uint16_t reason, const unsigned char *data, size_t length);
```
mpolitzer, line 292: It would be nice to have the other way as well, something like this:
```
CM_API int cm_recv_cmio_request(cm_machine *m, uint16_t *reason,
    size_t max_length, unsigned char *data, size_t *length);
```

edubart: I agree, I think this will be useful.

diegonehab: We don’t implement this on the solidity side. So we don’t know what order they will need things in the log. We are also going to simplify this protocol soon, right?

mpolitzer: I see this as an abstraction of the HTIF + buffers interface we have today. The alternative is for users to pack / unpack HTIF by themselves, not ideal.

> We are also going to simplify this protocol soon, right?

I'm not following. Is this about uarch -> risc0 transition?
if so we could drop _log_ and _uarch_, but Node people will still need to load / retrieve bytes from buffers right?

diegonehab: https://github.com/cartesi/machine-emulator/issues/257

diegonehab: We can add a method for the reader node to use. But we won’t have a method for the verifier node to use. This would return a log. But we don’t control the blockchain implementation of whatever it is that would consume that log. So in what would be the information is it that we would provide ind the log and in what order?

edubart: I think this could be useful on its own as a high level API to do IO with the machine through the C API for applications that want to use libcartesi as a C library. As of now after the machine yields you can send data inside with `cm_send_cmio_response` but there is no easy function to read data from inside, `cm_recv_cmio_request` would be this function.

diegonehab: Go for it.

mpolitzer: > #257 

So not having to deal with `iflags`, only HTIF. That is how 

> So in what would be the information is it that we would provide ind the log and in what order?

Assuming: "So what would be the information we provide, and in what order?"

A bit out of my depth, but I don't see the issue. My understanding is that this is already a provable machine since it is the machine immediately before a surgery.

So for instance, we do `read_htif(...);` + `read_memory(TX_BUFFER);` in `cartesi-machine.lua`.
What is the difference from that to having a single function that does both and massages the internals?

diegonehab: Because we don’t control what this code will look like in solidity. We don’t know anything about the different domains.

diegonehab: The log has to include the places where the info comes from. So the changes impact this.

src/machine-c-api.h
```diff
@@ -82,1894 +83,528 @@ typedef enum { // NOLINT(modernize-use-using)
...
+/// \param config Machine configuration as a JSON string.
+/// \param runtime_config Machine runtime configuration as a JSON string.
+/// \param new_machine Receives the pointer to new machine instance.
+/// \returns 0 for success, non zero code for error.
+CM_API int cm_create(const char *config, const char *runtime_config, cm_machine **new_machine);
+
+/// \brief Destroys a machine.
+/// \returns 0 for success, non zero code for error.
+/// \details The machine is deallocated and its pointer cannot be used after this call.
+CM_API int cm_destroy(cm_machine *m);
```
mpolitzer, line 191: We can set the pointer back to NULL depending on how paranoid you want to be.
```
CM_API int cm_destroy(cm_machine **m);
```

src/machine-c-api.h
```diff
@@ -82,1894 +83,528 @@ typedef enum { // NOLINT(modernize-use-using)
...
+    CM_CSR_IFLAGS_Y,           ///< View of Y field from IFLAGS
+    CM_CSR_IFLAGS_H,           ///< View of H field from IFLAGS
+} CM_CSR;
+
+/// \brief Machine instance handle.
+/// \details cm_machine* is handle used to pass the machine object when calling C API functions.
+typedef struct cm_machine cm_machine;
+
+/// \brief Machine hash array.
+typedef uint8_t cm_hash[CM_MACHINE_HASH_BYTE_SIZE];
```
mpolitzer, line 175: This can be error prone when used as a function argument, for instance.

Prefer something like this.
```
typedef struct {
    uint8_t data[CM_MACHINE_HASH_BYTE_SIZE];
} cm_hash;
```

src/machine-c-api.h
diegonehab: Meh.

edubart: Yea, not doing this hehe

diegonehab:
> Added views CSRs for HTIF fromhost/tohost for dev/data/cmd (eg. CM_CSR_HTIF_TOHOST_DEV), the idea is to write on fields of these CSRs through cm_write_csr/cm_read_csr.

Should we have the REASON field there as well? And make DATA smaller? It's a question I've always had...

edubart:
> Should we have the REASON field there as well? And make DATA smaller? It's a question I've always had...

I think we should, I went ahead I did the change on the branch with the actual changes I am working on, 4 bytes for data, 2 bytes for reason, this simplified some fragile manual bit shifting in some lua scripts.

edubart:
The API mentioned in this PR was implemented at https://github.com/cartesi/machine-emulator/pull/273 and the work continues there, I am closing this PR to avoid confusion.

PR #254, opened 2024-07-02
Feature/output unification v2

mpolitzer:
### TODO:
- [x] wait for the actual tools release. (needs commit: https://github.com/cartesi/machine-emulator-tools/pull/71)

## commits

- feat: update test data due to prev_randao field

- feat: output unification v2
    
    Outputs root hash merkle tree is the same for all inputs, we stop
    resetting it for each input (v1). With that the tree will grow
    indefinetely and thus we also increased its height to support 2^63 leafs.

- fix: remove invalid test case
    
    The machine itself does not take any steps to restore its state, the
    the emulator driver / controller does. As such we cannot expect the
    machine to reset its own state and thus this test case makes no sense.


vfusco (changes requested):
Tools v0.16.0 is already released. Please update the references.

vfusco (approved):

edubart (approved):

PR #262, opened 2024-08-08
fix: wait for gRPC to completely shutdown before forking

edubart:


src/remote-machine.cpp
```diff
@@ -1496,6 +1501,18 @@ static void server_loop(const char *server_address, const char *session_id, cons
         hctx.s.reset(nullptr);
         hctx.cq.reset(nullptr);
 
+        // Being shutting-down gRPC asynchronously and decrease its initialization counter
+        grpc_shutdown();
+        // Wait gRPC to completely shutdown
+        auto grpc_shutdown_deadline = std::chrono::system_clock::now() + std::chrono::milliseconds(shutdown_deadline);
```
mpolitzer, line 1507: Consider using `steady_clock` instead to avoid clock adjustment shenanigans.

edubart:
The release of v0.16.2 is not in the plans anymore, so I am closing this PR. Whoever is facing this gRPC bug can usually workaround the issue by running the node in bigger machines (more CPUs and RAM).

PR #266, opened 2024-08-26
fix: avoid extra root hash computation when storing an image snapshot

edubart:
I found out while investigating why a "reader node" using the CLI was too slow, that `--store` is always requesting a root hash computation when saving a snapshot, despite not using it, making the store operation slow, and this PR fixes the issue.


mpernambuco (approved):
good catch

mpolitzer (approved):

PR #275, opened 2024-09-17
feat: return pid in jsonrpc remote machine fork

edubart:


edubart:
This was included in https://github.com/cartesi/machine-emulator/pull/273, closing.

PR #284, opened 2024-10-21
fix: HTIF_YIELD_MANUAL is nil, this blocks cli assert-rolling-template

stskeeps:
- When having a valid rolling template, assert_rolling_template always fails. This is because of it using HTIF_YIELD_MANUAL for comparison which doesn't exist. It should be HTIF_YIELD_CMD_MANUAL.

### DIFF
--- a/src/cartesi-machine.lua
+++ b/src/cartesi-machine.lua
@@ -2319,7 +2319,7 @@ if assert_rolling_template then
     local cmd, reason = get_yield(machine)
     if
         not (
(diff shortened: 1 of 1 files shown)


edubart:
Thanks for spotting and reporting this, this was fixed during the C API refactoring that was merged today. Due to the renaming and refactoring that has occurred, this constant name is now actually `CMIO_YIELD_COMMAND_MANUAL`, so this PR is no longer applicable.

PR #297, opened 2024-12-17
feat!: add log_step and verify_step

mpernambuco:


.github/workflows/build.yml
```diff
@@ -191,6 +191,9 @@ jobs:
       - name: Run test suite inside the docker image
         run: docker run --rm -t ${{ github.repository_owner }}/machine-emulator:tests /usr/bin/cartesi-machine-tests run
 
+      - name: Run test suite with log_stepinside the docker image
```
diegonehab, line 194: Typo.

edubart (changes requested):

src/cartesi-machine.lua
```diff
@@ -1251,6 +1256,15 @@ local options = {
             return true
         end,
     },
+    {
+        "^%-%-log%-step%=(.*),(.*)$",
+        function(count, filename)
+            if (not count) or not filename then return false end
```
diegonehab, line 1262: Are we still running format-lua?

src/clua-i-virtual-machine.cpp
```diff
@@ -1048,6 +1058,19 @@ static int machine_obj_index_get_reg_address(lua_State *L) {
...
 }
 
+/// \brief This is the machine.verify_step() method implementation.
+static int machine_obj_index_verify_step(lua_State *L) {
+    lua_settop(L, 4);
+    cm_hash root_hash_before{};
+    clua_check_cm_hash(L, 1, &root_hash_before);
+    cm_hash root_hash_after{};
+    clua_check_cm_hash(L, 4, &root_hash_after);
+    if (cm_verify_step(&root_hash_before, luaL_checkstring(L, 2), luaL_checkinteger(L, 3), &root_hash_after) != 0) {
```
diegonehab, line 1068: We need this to go through the machine pointer as well. See for example implementation of `cm_verify_step_uarch(const cm_machine *m, const cm_hash *root_hash_before, const char *log, const cm_hash *root_hash_after)`.

src/jsonrpc-discover.json
```diff
@@ -841,6 +841,80 @@
...
+        "name": "reason",
+        "description": "Reason call returned",
+        "schema": {
+          "$ref": "#/components/schemas/InterpreterBreakReason"
+        }
+      }
+    },
+    {
+      "name": "machine.verify_step",
+      "summary": "Checks the validity of a step log fil.",
```
diegonehab, line 876: Typo.

src/machine-c-api.cpp
```diff
@@ -352,6 +364,19 @@ cm_error cm_log_step_uarch(cm_machine *m, int32_t log_type, const char **log) tr
...
 }
 
+cm_error cm_verify_step(const cm_hash *root_hash_before, const char *log_filename, uint64_t mcycle_count,
+    const cm_hash *root_hash_after) try {
+    if (log_filename == nullptr) {
+        throw std::invalid_argument("invalid log_filename");
+    }
+    const cartesi::machine::hash_type cpp_root_hash_before = convert_from_c(root_hash_before);
+    const cartesi::machine::hash_type cpp_root_hash_after = convert_from_c(root_hash_after);
+    cartesi::machine::verify_step(cpp_root_hash_before, log_filename, mcycle_count, cpp_root_hash_after);
```
diegonehab, line 374: This has to receive a `const cm_machine *m`. If it is nullptr, then use the static method directly. If it is not null, convert to `cpp_m` and call `cpp_m->verify_step()`

src/machine-c-api.h
```diff
@@ -704,6 +713,19 @@ CM_API cm_error cm_log_send_cmio_response(cm_machine *m, uint16_t reason, const
 // Verifying
 // ------------------------------------
 
+/// \brief Checks the validity of a step log file.
```
diegonehab, line 716: Update help with new `const cm_machine *m` argument.

tests/lua/machine-bind.lua
```diff
@@ -540,13 +540,13 @@ end)
 
 do_test("should error if target mcycle is smaller than current mcycle", function(machine)
     machine:write_reg("mcycle", MAX_MCYCLE)
-    assert(machine:read_mcycle() == MAX_MCYCLE)
+    assert(machine:read_reg("mcycle") == MAX_MCYCLE)
```
edubart, line 543: We still have `read_mcycle`, no need to change to `read_reg` in this case and all the ones below.

tests/lua/machine-bind.lua
```diff
@@ -1497,4 +1497,206 @@ test_util.make_do_test(build_machine, machine_type, {
     assert(log.accesses[7].written == leaf_data)
 end)
 
+-- helper function to load a step log file into a table
+local function read_step_log_file(filename)
+    local file <close> = io.open(filename, "rb")
```
edubart, line 1502: Wrap `io.open` with `assert()`, here in all new `io.open` below, otherwise my Lua linter will complain.

src/machine-c-api.cpp
```diff
@@ -336,6 +336,18 @@ cm_error cm_run_uarch(cm_machine *m, uint64_t uarch_cycle_end, cm_uarch_break_re
     return cm_result_failure();
 }
 
+CM_API cm_error cm_log_step(cm_machine *m, uint64_t mcycle_count, const char *log_filename,
+    cm_break_reason *break_reason) try {
+    auto *cpp_m = convert_from_c(m);
+    const auto status = cpp_m->log_step(mcycle_count, log_filename);
```
edubart, line 342: Consider the case where `log_filename` is invalid (e.g NULL).

src/machine-merkle-tree.cpp
```diff
@@ -431,6 +431,34 @@ machine_merkle_tree::proof_type machine_merkle_tree::get_proof(address_type targ
...
+        const int log2_child_size = log2_node_size - 1;
+        const int path_bit = static_cast<const int>((target_address & (UINT64_C(1) << (log2_child_size))) != 0);
+        node = node->child[path_bit];
+        log2_node_size = log2_child_size;
+    }
+    if (node == nullptr) {
+        // We hit a pristine node along the path to the target node
+        return get_pristine_hash(log2_target_size);
+    }
+    assert(node);
```
edubart, line 458: Unnecessary `assert` since `node == nullptr` was checked previously.

src/replay-step-state-access-interop.cpp
```diff
@@ -0,0 +1,46 @@
...
+    interop_hash_type result) {
+    machine_merkle_tree::hasher_type hasher{};
+    // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast)
+    get_concat_hash(hasher, *reinterpret_cast<const machine_merkle_tree::hash_type *>(left),
+        *reinterpret_cast<const machine_merkle_tree::hash_type *>(right),
+        *reinterpret_cast<machine_merkle_tree::hash_type *>(result));
+    // NOLINTEND(cppcoreguidelines-pro-type-reinterpret-cast)
+}
+
+extern "C" void interop_print(const char *msg) {
```
edubart, line 44: I don't think this function is being used, could it be removed?

edubart (changes requested):

src/record-step-state-access.h
```diff
@@ -0,0 +1,840 @@
...
+        (void) type;
+        (void) text;
+    }
+
+    int do_make_scoped_note(const char *text) { // NOLINT(readability-convert-member-functions-to-static)
+        (void) text;
+        return 0;
+    }
+
+    uint64_t do_read_x(int reg) const {
```
diegonehab, line 173: Had a conversation with @mpernambuco and I think we should replace all of these with `do_read_reg()`/`do_write_reg()`. Then, we implement all `i_state_access::do_read_*.*/do_write*.*` on top of `i_state_access::read_reg()/write_reg()`, and we forward `i_state_access::read_reg()/write_reg()` to `i_state_access::do_read_reg()/do_write_reg()` and implement `i_state_access::do_read_reg()/do_write_reg()` to throw and exception. This way, every state access can choose to implement `do_read_*.*()/do_write_*.*()` or only implement `do_read_reg()` and `do_write_reg()`, or even mix both. What do you think, @edubart? We should also merge the enums `shadow_state_reg` with `machine::reg`, so we don't need to convert between them anymore.

diegonehab: Maybe do this in a separate PR?

src/record-step-state-access.h
```diff
@@ -0,0 +1,840 @@
...
+            }
+            // Otherwise, if we found an entry where the access fits, return it
+            // Note the "strange" order of arithmetic operations.
+            // This is to ensure there is no overflow.
+            // Since we know paddr >= start, there is no chance of overflow
+            // in the first subtraction.
+            // Since length is at least 4096 (an entire page), there is no
+            // chance of overflow in the second subtraction.
+            if (paddr >= pma.get_start() && paddr - pma.get_start() <= pma.get_length() - sizeof(T)) {
+                touch_page(paddr);
```
diegonehab, line 673: Why touch `paddr` itself here? Won't you touch it elsewhere afterwards?

src/replay-step-state-access.h
```diff
@@ -0,0 +1,1099 @@
...
+// \brief Provides machine state from a step log file
+class replay_step_state_access : public i_state_access<replay_step_state_access, mock_pma_entry> {
+public:
+    using hash_type = std::array<unsigned char, interop_machine_hash_byte_size>;
+    static_assert(sizeof(hash_type) == interop_machine_hash_byte_size);
+
+private:
+    using address_type = uint64_t;
+    using data_type = unsigned char[PMA_PAGE_SIZE];
+#pragma pack(push, 1)
```
diegonehab, line 174: Use `PACKED` from `compiler-defines.h`.

src/replay-step-state-access.h
```diff
@@ -0,0 +1,1099 @@
...
+        }
+    }
+
+private:
+    friend i_state_access<replay_step_state_access, mock_pma_entry>;
+
+    // \brief Relocate all TLB virtual to host offsets
+    // \details Points the vh_offset relative to the logged page data
+    template <TLB_entry_type ETYPE>
+    void relocate_all_tlb_vh_offset() {
```
diegonehab, line 293: I'll probably need some help to go over the TLB tricks.

tests/lua/machine-bind.lua
```diff
@@ -1497,4 +1497,206 @@ test_util.make_do_test(build_machine, machine_type, {
...
+    -- get current root hash and log step
+    local root_hash_before = machine:get_root_hash()
+    local mcycle_count = 10
+    local status = machine:log_step(mcycle_count, filename1)
+    assert(status == cartesi.BREAK_REASON_REACHED_TARGET_MCYCLE)
+    assert(machine:read_reg("mcycle") == mcycle_count)
+    local root_hash_after = machine:get_root_hash()
+    assert(root_hash_before ~= root_hash_after)
+    -- verify step should pass
+    machine.verify_step(root_hash_before, filename1, mcycle_count, root_hash_after)
```
diegonehab, line 1582: machine:verify_step()

diegonehab (changes requested):
See comments.

src/machine-c-api.h
```diff
@@ -672,6 +672,15 @@ CM_API cm_error cm_send_cmio_response(cm_machine *m, uint16_t reason, const uint
 // Logging
 // ------------------------------------
 
+/// \brief Runs the machine for the given mcycle count and generates a log of accessed pages and proof data.
+/// \param m Pointer to a non-empty machine object (holds a machine instance).
+/// \param mcycle_count Number of mcycles to run
+/// \param log_filename Name of the log file to be generated
+/// \param break_reason Receives reason for returning (can be NULL).
+/// \returns 0 for success, non zero code for error.
+CM_API cm_error cm_log_step(cm_machine *m, uint64_t mcycle_count, const char *log_filename,
```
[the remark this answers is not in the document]
edubart: Never mind, this could come as enhanced in the future if there is need.

src/record-step-state-access.h
edubart: I agree, I think this would a good to simplify all state accessors, but indeed this refactoring deserves a separate PR.

src/cartesi-machine.lua
mpernambuco: yes

src/record-step-state-access.h
mpernambuco: It is not needed. Good catch.

src/machine.cpp
```diff
@@ -2482,6 +2496,39 @@ uarch_interpreter_break_reason machine::run_uarch(uint64_t uarch_cycle_end) {
...
+void machine::verify_step(const hash_type &root_hash_before, const std::string &filename, uint64_t mcycle_count,
+    const hash_type &root_hash_after) {
+    auto data_length = os_get_file_length(filename.c_str(), "step log file");
+    auto *data = os_map_file(filename.c_str(), data_length, false /* not shared */);
+    replay_step_state_access a(data, data_length, root_hash_before);
+    uint64_t mcycle_end{};
+    if (__builtin_add_overflow(a.read_mcycle(), mcycle_count, &mcycle_end)) {
+        mcycle_end = UINT64_MAX;
+    }
+    interpret(a, mcycle_end);
```
edubart, line 2527: Why `break_reason` is not being returned? I think it wouldn't hurt to return it, and could be useful in some cases.

src/cartesi-machine.lua
```diff
@@ -2185,6 +2199,12 @@ if max_uarch_cycle > 0 then
     end
 end
 if gdb_stub then gdb_stub:close() end
+if log_step_mcycle_count then
+    stderr(string.format("Logging step of %d cycles to %s\n", log_step_mcycle_count, log_step_filename))
+    print_root_hash(machine, stderr_unsilenceable)
```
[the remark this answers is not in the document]
mpernambuco: We need to print the hash before the call to log_step() because it occurs after the machine may have advanced a certain number of cycles. Similarly, printing the hash after is necessary because the machine can advance micro cycles following the logging of the step.

The --initial-hash and --final-hash flags only print the hashes of the very first and very last cycles.

It is reasonable to assume that users saving a step log would also want the corresponding hashes to validate the log elsewhere, hence the prints.

For example:
```
cartesi-machine.lua --initial-hash --max-mcycle=10 --log-step=1,/tmp/loxa1.bin --max-uarch-cycle=1 --final-hash
0: da95ad8352d6ff9183c8b3f7520c6a68097f42ef42ff3ddb52eb2238d325aad3
Logging step of 1 cycles to /tmp/loxa1.bin
10: ef886a04bc473480929ec2d14608286bf2705ed9a8767edcd30299131afc9f6f
11: b4182e651d2fa9036f255fc26c946d8432bd3b4ab50790415897e6d411b30a73
11: b4182e651d2fa9036f255fc26c946d8432bd3b4ab50790415897e6d411b30a73
```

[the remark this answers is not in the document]
diegonehab: Hmmm. I we should throw if the uarch is not in a reset state when we do log-step. Mixing these two is basically a bug, right? Or some very adventurous person doing something way over their head. Can we do that?

tests/lua/cartesi-machine-tests.lua
```diff
@@ -586,6 +589,22 @@ local function run_machine(machine, ctx, max_mcycle, advance_machine_fn)
     ctx.read_htif_tohost_data = machine:read_reg("htif_tohost_data")
 end
 
+local function run_machine_step(machine, ctx, mcycle_count)
```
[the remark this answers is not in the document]
mpernambuco: Good idea.

src/cartesi-machine.lua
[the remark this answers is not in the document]
mpernambuco: `log_step_uarch()` is always the last advance operation, when `--log-step-uarch`  is specified.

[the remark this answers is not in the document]
diegonehab: I mean this combination `--log-step=1,/tmp/loxa1.bin --max-uarch-cycle=1` should not be acceptable.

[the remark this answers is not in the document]
mpernambuco: Got it.
The micro steps  need to occur after all big steps.
```
cartesi-machine.lua --initial-hash --max-mcycle=10 --log-step=1,/tmp/loxa1.bin --max-uarch-cycle=8000 --log-step-uarch --final-hash
0: da95ad8352d6ff9183c8b3f7520c6a68097f42ef42ff3ddb52eb2238d325aad3
Logging step of 1 cycles to /tmp/loxa1.bin
10: e9d613431892a7c6d4fafb64e828bf772c9583081de6fd8141943792909dd94d
11: 763ccef8428a0f68356ff421cdfcd4da2832160dd41fdac7d245c016b323424b
Cycles: 12
uCycles: 262
Gathering micro step log: please wait
begin step
  1: read uarch.cycle@0x400008(4194312): 0x106(262)
  2: read uarch.halt_flag@0x400000(4194304): 0x1(1)
end step
12: c75e0fd7e098ba66baf20d0baa2ec7a2452003431334b587d7ccc5079c613e49
``

edubart (approved):

diegonehab (approved):

PR #299, opened 2024-12-23
Split iflags into 4 registers

diegonehab:
This PR splits the iflags register in to 4 64-bit registers: iprv, iflags_X, iflags_Y, iflags_H.
This simplifies the implementation and paves the way for a lot of further simplifications because we do not have to implement read/modify/write when accessing pieces of a single 64-bit register anymore.
This even reduces the number of accesses in the access-log for send_cmio_response.
The PR goes further and removes the special register access functions from the C and Lua API.
Now every access must go through cm_read_reg/cm_write_reg.


edubart (approved):
I've tested and reviewed emulator side of things, and I approve.

PR #306, opened 2025-01-20
Bump clang-tidy to 19 and fix linter errors

edubart:


.clang-tidy
```diff
@@ -44,7 +43,6 @@ Checks: >-
   -readability-function-cognitive-complexity,
   -readability-identifier-length,
   -readability-magic-numbers,
-  -readability-enum-initial-value,
```
diegonehab: I don't really like this. The standard is pretty clear about what the values would be. So no need to enforce use writing +1, +2 etc. Really, I have no idea why somebody thinks this is less error prone.

src/virtio-p9fs.h
```diff
@@ -55,73 +55,73 @@ enum virtio_p9fs_constants : uint32_t {
 /// \brief 9P2000.L opcodes
 enum p9_opcode : uint8_t {
     P9_TLERROR = 6,
```
diegonehab, line 57: This here is being lazy. :)

src/machine-reg.h
```diff
@@ -177,15 +177,15 @@ enum class machine_reg : uint64_t {
     uarch_last_ = uarch_x31,
 
     // Views of registers
-    htif_tohost_dev,
```
diegonehab, line 180: How do you know PMA_SHADOW_STATE_LENGTH+8 doesn't fall inside PMA_SHADOW_UARCH_STATE? See? The compiler will not warn about repeated enums, because we use repeated enums all the time.

src/machine.cpp
```diff
@@ -795,7 +795,7 @@ const pma_entry &machine::find_pma_entry(const CONTAINER &pmas, uint64_t paddr,
 
 template <typename T>
```
diegonehab, line 788: I want to get rid of this entirely. I want to remove machine::find_pma_entry altogether, and have the update_merkle_tree_page or whatever that uses it t o simply use std::find on the m_merkle_pmas array.

diegonehab (changes requested):
I just don't like the enum one.

diegonehab (approved):

PR #307, opened 2025-01-24
fix: fix inconsistencies in jsonrpc-discover.json

edubart:


edubart:
Since CI for arm64 started failing, I included this in https://github.com/cartesi/machine-emulator/pull/296

PR #313, opened 2025-03-21
Fix record_send_cmio_response::do_write_memory_with_padding

diegonehab:
When writing to the underlying PMA, the code was ignoring the offset between paddr and the PMA start.
I believe that, in practice, since we always invoke this function using paddr == rx_buffer.get_start(), the problem doesn't ever manifest. Still, it is a bug waiting to happen.


mpernambuco (dismissed):
Good catch!

edubart (approved):

PR #320, opened 2025-04-04
fix: conflicting symbols between of cartesi/jsonrpc libraries

edubart:


edubart:
This was merged in https://github.com/cartesi/machine-emulator/pull/318

PR #322, opened 2025-04-10
Refactor/penumbra

edubart:
The goal of this PR is to add support for backing files to shadow address ranges of machines, so we can run machines directly mapped from disk:
- [x] shadow uarch state
- [ ] shadow processor
- [ ] shadow tlb

It also adds a "penumbra" region that only exists during runtime to the end of the machine state, however it's not stored in disk, in order to keep the interpreter efficient.


edubart:
this was superseded by https://github.com/cartesi/machine-emulator/pull/333

PR #327, opened 2025-05-26
Add configurable hash tree target support

mpernambuco:
This PR introduces a new machine configuration setting - `hash_tree_target`.
It determines if the hash tree will be used by the uarch or RISC Zero environments.
The hashing algorithm changes depending on the `hash_tree_target`; uarch uses Keccak, RISC Zero uses SHA-256.
The scoped/templated hash_types have been replaced by machine_hash; always 32 bytes long.
i_hasher is no longer a CRTP interface. It is now a class that can do either Keccak or SHA-256. The name is still i_hash to make it easier to do rename and merge with the upcoming hash tree simplification/optimization.


edubart:
I am closing this because this some portions of it was adopted and adjusted to support vectorized SHA-256 hashing in https://github.com/cartesi/machine-emulator/pull/335

PR #332, opened 2025-07-24
Parallel vectorized Keccak-256/SHA-256 merkle tree hashing

edubart:
This PR:
- Adds custom Keccak-256/SHA-256 implementation using SIMD instructions optimized for hash operations we use in our hash tree (with fixed size leaves/nodes), it uses compiler vector extensions (GCC/Clang compatible) for a simpler implementation (without compiler intrinsics).
- Adds specialized code paths leveraging AVX2 and AVX512 instructions on x86_64 using GCC function multiversioning
- Optimizes pristine page checking with better SIMD instruction utilization
- Makes hash function configurable (adapted from #327) - defaults to Keccak-256, can be set to SHA-256 via `--hash-tree=hash_function:sha256`
- Adds new C API hash functions (`cm_get_hash` and `cm_get_concat_hash`)
- Adds Lua hash functions `cartesi.keccak256` (renamed from `cartesi.kecccak`) and `cartesi.sha256` using the new C APIs
- Optimizes hash tree multithreading by reducing atomic operations and improving scheduling

The overall speed boost of periodic sparse hashing at 32K cycles relative to base branch (refactor/new-merkle-tree) is more than 2x in average for must benchmark I run (on my hardware supporting only AVX2 instructions).

### DIFF
--- a/src/clua-cartesi.cpp
+++ b/src/clua-cartesi.cpp
@@ -31,12 +31,10 @@ extern "C" {
 #include "clua-i-machine.h"
 #include "clua.h"
 #include "htif-constants.h"
-#include "keccak-256-hasher.h"
 #include "machine-c-api.h"
 #include "machine-c-version.h"
 #include "machine-hash.h"
 #include "riscv-constants.h"
-#include "sha-256-hasher.h"
 #include "uarch-constants.h"
 #include "uarch-pristine.h"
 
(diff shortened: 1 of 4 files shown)


edubart:
this was superseded by https://github.com/cartesi/machine-emulator/pull/333

Issue #314, opened 2025-03-26
Provide C API to get version at runtime


edubart:
## Context

Sometimes we may want to check the emulator version through the C API at runtime (not at compile time), unfortunately there is no C API for that. A workaround is to instantiate a new machine, then check its MARCHID CSR, but it's a dirty solution.

This problem became more relevant after the Dave team asked ways to make sure they are running the correct emulator version during fraud proofs with machine-solidity-step.

## Possible solutions

We could add a `cm_get_version` to return emulator semantic version (major, minor, patch, label), plus `MARCHID`, `MIMPID` at runtime. Maybe add all values from `cartesi-machine --version-json`. Pristine uarch RAM hash would also be nice, so we could check in `machine-solidity-step` in Dave to make sure the uarch pristine RAM matches before engaging in fraud proofs.

Issue #311, opened 2025-02-11
Support stdio flush runtime option


edubart:
## Context

When running heavy console applications in non interactive mode (e.g a heavy LLM which can take hours to process) the stdout is flushed only on every new line, meaning it's impossible to see intermediary sentence output tokens being generated lively, ideally we want let users hacking to see what is being generated in the console.

## Possible solutions

We could add new runtime option to set stdout flushing scheme (e,g every new line, every character, or never flush).

Issue #276, opened 2024-10-03
Redirection of `stdout`


edubart:
## Context

For some use cases is useful to to spawn multiple machines in different threads and yet still be able to capture stdout for each machine separately. We need to find a way to do this changing at runtime how `os_putchar` behaves.

## Possible solutions

There are a couple of ways of doing this:
1. A callback (kinda unsafe and harder to use when binding to other languages)
2. A machine run interruption with a new break reason with some buffering (perhaps the best option, more elaborate to do it)
3. A runtime option to set the file descriptor number os_putchar uses (easy/quick to do, not that cross platform, still need to create FD for each machine)

We are inclined to support option 2.

PR #337, opened 2025-11-06
Support for machine console redirection

edubart:
This PR solves:
- https://github.com/cartesi/machine-emulator/issues/311
- https://github.com/cartesi/machine-emulator/issues/276
- https://github.com/cartesi/machine-emulator/issues/314


src/htif-address-range.cpp
```diff
@@ -89,11 +89,14 @@ static execute_status htif_yield(i_device_state_access *a, uint64_t cmd, uint64_
...
 
 static execute_status htif_console(i_device_state_access *a, uint64_t cmd, uint64_t data) {
+    auto status = execute_status::success;
     // If console command is enabled, perform it and acknowledge
     if (cmd < 64 && (((a->read_htif_iconsole() >> cmd) & 1) != 0)) {
         if (cmd == HTIF_CONSOLE_CMD_PUTCHAR) {
             const uint8_t ch = data & 0xff;
-            a->putchar(ch);
+            if (a->putchar(ch)) {
+                status = execute_status::success_and_flush_console;
```
diegonehab: I see now that the break reason may not have the buffer full, so my suggestion for name isn't great. Yours is a verb here. I prefer it to the "flushed console" that you used. How about CM_BREAK_REASON_CONSOLE_OUTPUT and CM_BREAK_REASON_CONSOLE_INPUT?

src/jsonrpc-machine.cpp
```diff
@@ -752,6 +752,29 @@ uint64_t jsonrpc_machine::do_translate_virtual_address(uint64_t vaddr) {
     return result;
 }
 
+uint64_t jsonrpc_machine::do_read_console_output(uint8_t *data, uint64_t max_length) {
+    uint64_t read_len{0};
+    if (max_length > 0) {
+        std::string result;
+        request("machine.read_console_output", std::tie(max_length), result);
+        std::string bin = cartesi::decode_base64(result);
+        std::memcpy(data, bin.data(), bin.length()); // NOLINT(bugprone-not-null-terminated-result)
```
diegonehab: I'd check `bin.length()` against `max_length` to be safe.

src/machine-console.cpp
```diff
@@ -0,0 +1,538 @@
...
+        os::close_fd(m_input_fd);
+        m_input_fd = -1;
+    }
+    m_input_file.reset();
+    m_input_filename.clear();
+    m_input_source = console_input_source::from_null;
+    m_input_closed = true;
+}
+
+void machine_console::validate_config(const console_runtime_config &config) {
```
diegonehab: Ah, so you validate here. But in the command-line cartesi-machine.lua you probably need to validate differently, right?

src/cartesi-machine.lua
```diff
@@ -1217,6 +1284,44 @@ local options = {
         "^%-it$",
         handle_interactive,
     },
+    {
```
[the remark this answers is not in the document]
diegonehab: But in Lua you change the flags based on which options have been set and you pick a preferred order. If two of these options are set, the user screwed up, no?

src/htif-address-range.cpp
diegonehab: It is unclear if we should treat console output as something "serious" or not. Shall we ask @vfusco and @tuler and somebody else?

src/machine-console.cpp
```diff
@@ -0,0 +1,538 @@
...
+                consumed_bytes = -1;
+            }
+            break;
+        case console_output_destination::to_buffer:
+            // Nothing to do, user must consume the buffer.
+            break;
+    }
+
+    if (consumed_bytes < 0) {
+        if (errno == EPIPE) { // Output was closed externally
```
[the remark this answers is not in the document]
edubart: >  The fwrite() function returns the number of members successfully written, which may be less than nitems if a write error is encountered. If size or nitems is 0, fwrite() returns 0 and the state of the stream remains unchanged. Otherwise, if a write error occurs, the error indicator for the stream is set and errno is set to indicate the error. 
> https://pubs.opengroup.org/onlinepubs/7908799/xsh/fwrite.html

src/jsonrpc-discover.json
```diff
@@ -735,6 +735,59 @@
         }
       }
     },
+    {
+      "name": "machine.get_console_output_length",
```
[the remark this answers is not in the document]
edubart: This method was removed.

src/jsonrpc-machine.cpp
edubart: Nice catch, I added a check for this.

src/jsonrpc-machine.cpp
```diff
@@ -752,6 +752,36 @@ uint64_t jsonrpc_machine::do_translate_virtual_address(uint64_t vaddr) {
...
+        }
+        std::memcpy(data, bin.data(), bin.length()); // NOLINT(bugprone-not-null-terminated-result)
+        read_len = bin.length();
+    }
+    return read_len;
+}
+
+uint64_t jsonrpc_machine::do_write_console_input(const uint8_t *data, uint64_t length) {
+    uint64_t result = 0;
+    if (data == nullptr && length == 0) { // Query mode
```
diegonehab: || ?

edubart: Not really, both must be true to be in query mode. If I do this change some tests would break. When you do `machine:write_console_input("")` in Lua, it calls this function with `data != nullptr` and `length == 0`, which is valid and then nothing will be written. While calling `machine:write_console_input()` will call this function with `data == nullptr` and `length == 0`, then this will be the query mode.

So I designed the APIs to be only in query mode if both conditions are true, using `||` will have some unwanted consequences, which would need API design changes.

src/machine.cpp
```diff
@@ -1927,10 +1937,33 @@ uarch_interpreter_break_reason machine::run_uarch(uint64_t uarch_cycle_end) {
         throw std::runtime_error("microarchitecture cannot be used with unreproducible machines");
     }
     const uarch_state_access a(*this);
-    os_silence_putchar(m_r.htif.no_console_putchar);
     return uarch_interpret(a, uarch_cycle_end);
 }
 
+template <typename STATE_ACCESS>
```
diegonehab: Nice and symmetrical. :)

src/jsonrpc-machine.cpp
edubart: > Do you throw and you have nullptr and non-zero length?

Yes in both `machine_console::read_output` and `machine_console::write_input`.

src/machine-c-api.h
```diff
@@ -645,6 +652,30 @@ CM_API cm_error cm_write_virtual_memory(cm_machine *m, uint64_t address, const u
...
+/// \details Requires the runtime console output destination to be set to buffer.
+/// Reads up to \p max_length bytes, or all available data if less is available.
+/// Read data is consumed and removed from the buffer.
+CM_API cm_error cm_read_console_output(cm_machine *m, uint8_t *data, uint64_t max_length, uint64_t *read_len);
+
+/// \brief Writes data to the console input buffer.
+/// \param m Pointer to a non-empty machine object (holds a machine instance).
+/// \param data Buffer containing data to write. Must be at least \p length bytes long.
+/// \param length Number of bytes to write.
+/// If zero and \p data is NULL, \p written_len receives the available space size that could be written.
```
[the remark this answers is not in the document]
edubart: I did that way at first but it had some unwanted side effects, like `machine:write_console_input()` doing the same as `machine:write_console_input("")`, but they are are different prototypes, and I thought it was odd for both to behave as in query mode.

src/machine-console.cpp
```diff
@@ -0,0 +1,538 @@
...
+    if (m_input_size == 0) {
+        m_input_start = 0;
+    }
+    return chunk_buf;
+}
+
+std::span<const uint8_t> machine_console::append_output(std::span<const uint8_t> buf) noexcept {
+    // Make room by discarding oldest data if the output buffer is too small
+    if (m_output_size + buf.size() > m_output_buffer.size()) {
+        std::ignore = std::fprintf(stderr,
```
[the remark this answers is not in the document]
diegonehab: Let’s forbid buffers smaller than 2048. What do you think?

PR #380, opened 2026-04-17
Implement the new NVRAM option

diegonehab:
This is an implementation of issue #370.


src/interpret.cpp
```diff
@@ -794,6 +794,20 @@ static FORCE_INLINE int32_t insn_get_C_LW_C_SW_imm(uint32_t insn) {
     return static_cast<int32_t>(((insn >> (10 - 3)) & 0x38) | ((insn >> (6 - 2)) & 0x4) | ((insn << (6 - 5)) & 0x40));
 }
 
+/// \brief Obtains the immediate value from C.LBU and C.SB instructions (Zcb extension).
+/// \param insn Instruction.
+/// \details This function is forced to be inline because GCC may not always inline it.
+static FORCE_INLINE uint32_t insn_get_C_LS_B_uimm(uint32_t insn) {
+    return ((insn >> 6) & 0b1) | ((insn >> 4) & 0b10);
```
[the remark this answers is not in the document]
edubart: No, it should be reversed according to the RISC-V spec, it is a special case.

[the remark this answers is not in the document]
diegonehab: It’s complaining about an inconsistency between the two places. Is it confused?

diegonehab:
Since the PR, the branch has accumulated a dozen more changes, grouped roughly as follows.

## New API surface

`get_address_name(paddr)` was added to the machine API (Lua, C, and JSON-RPC). It is the inverse of `get_reg_address`: given a physical address it returns the human-readable name of the register or memory region it belongs to (`"pc"`, `"tlb.slot.vaddr_page"`, `"pma.istart"`, `"memory"`, etc.). `--dump-memory-ranges` gained an optional directory argument so callers can redirect the output without a shell redirect.

## Refactoring

`machine_reg_get_name` was dropped (superseded by `get_address_name`). The uarch state-access helpers were narrowed — only the accessors actually needed by the uarch interpreter are instantiated, which trims dead code.

## Bug fixes

Label validation and argument checking were tightened for `replace_memory_range`.

## Test infrastructure

Coverage tracking was extended to Lua tests via a `luacov`→`gcov` bridge. `test-evmu` and `test-cm-cli` were converted from ad-hoc scripts to lester spec files (`spec-evmu.lua`, `spec-cm-cli.lua`).

## JSON-RPC coverage — new C test suite

A new `test-machine-c-jsonrpc-api.cpp` exercises the transport and lifecycle paths that have no analogue in the local test suite: argument validation, spawn/connect/disconnect, server version checking, cleanup modes (`SHUTDOWN` / `DESTROY` / `NOTHING`), and server-side "no machine" guards on every machine method.

## JSON-RPC coverage — remote run of the existing C API suite

The comprehensive `test-machine-c-api.cpp` is now parameterized: when `CARTESI_JSONRPC_TEST_REMOTE` is set it replays every existing test body against a spawned JSON-RPC server via thin `test_create_new` / `test_load_new` / `test_delete` wrappers, exercising all `machine.*` handlers and client-side wrappers over the wire without duplicating test logic. A `test-c-api-remote` make target drives this.

## JSON-RPC error propagation

To make the remote run pass, exception identity is now preserved across the wire. The `cm_error` enum was moved to its own header (`cm-error.h`). A new `cm-exception.hpp` provides a `noexcept` two-pass helper `cm_exception_to_error_code(string&)` (pass 1 extracts the message, pass 2 dispatches on dynamic type) and its reverse `cm_error_code_to_exception()`. The server encodes the `cm_error` integer in `error.data`; the client decodes it and re-throws the exact exception subclass, so `cm_result_failure()` returns the correct code on both sides.

`CM_ERROR_SYSTEM_ERROR` was dropped because `std::system_error` inherits from `runtime_error` and cannot be round-tripped portably — it maps to `CM_ERROR_RUNTIME_ERROR` on both paths. A null-buffer guard (`data == nullptr && length > 0`) was added to the public `i_machine` memory wrappers so all backends (local and remote) are protected at one point.

## Lua spec tests — remote variants

`spec-machine-api.lua`, `spec-hash-tree.lua`, and `spec-verify-step-failure.lua` each gained a remote variant that spawns a JSON-RPC server and runs the same assertions through the network, following the pattern established by `spec-collect-hashes.lua`. The few tests that are prohibitively slow over JSON-RPC (dense page hash iteration) are skipped in the remote variant.

src/cm-exception.hpp
```diff
@@ -0,0 +1,163 @@
...
+        return CM_ERROR_BAD_OPTIONAL_ACCESS;
+    } catch (const std::range_error &) {
+        return CM_ERROR_RANGE_ERROR;
+    } catch (const std::overflow_error &) {
+        return CM_ERROR_OVERFLOW_ERROR;
+    } catch (const std::underflow_error &) {
+        return CM_ERROR_UNDERFLOW_ERROR;
+    } catch (const std::regex_error &) {
+        return CM_ERROR_REGEX_ERROR;
+    } catch (const std::runtime_error &) {
```
edubart, line 93: Why the case for `std::system_error` was removed?

src/cm-error.h
```diff
@@ -0,0 +1,47 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License along
+// with this program (see COPYING). If not, see <https://www.gnu.org/licenses/>.
+//
+
+#ifndef CM_ERROR_H // NOLINTBEGIN
+#define CM_ERROR_H
+
+/// \brief Error codes returned from the C API.
+typedef enum cm_error {
```
edubart: I don't see the motivation to have this file with just one enum declaration, I would leave it in `cm.h` as it was before. Having `cm.h` self-contained makes life easier to create binding generators for `cm.h`. It's awkward to have all enums there except one.

src/cm.h
```diff
@@ -360,6 +337,14 @@ CM_API cm_error cm_get_default_config(const cm_machine *m, const char **config);
 /// \returns 0 for success, non zero code for error.
 CM_API cm_error cm_get_reg_address(const cm_machine *m, cm_reg reg, uint64_t *val);
 
+/// \brief Gets a description of what is at a given target physical address
```
edubart, line 384: An example would be helpful, I got confused if this would return the `label`, but it does not. I assume others could think this is the label.

src/dtb.cpp
```diff
@@ -171,22 +171,51 @@ void dtb_init(const machine_config &c, unsigned char *dtb_start, uint64_t dtb_le
             fdt.prop_u32("#address-cells", 2);
             fdt.prop_u32("#size-cells", 2);
             fdt.prop_empty("ranges");
-            { // reserve 256KB for firmware M-mode code (such as OpenSBI)
+            { // reserve 512KB for firmware M-mode code (such as OpenSBI)
                 fdt.begin_node_num("fw_resv", AR_RAM_START);
-                fdt.prop_u64_list<2>("reg", {AR_RAM_START, 0x40000});
+                fdt.prop_u64_list<2>("reg", {AR_RAM_START, 0x80000});
```
edubart: I guess we should start using a constant/define instead of hard-coding this.

src/dtb.cpp
```diff
@@ -171,22 +171,51 @@ void dtb_init(const machine_config &c, unsigned char *dtb_start, uint64_t dtb_le
...
+                fdt.prop_u64_list<2>("reg", {f.start, f.length});
+                fdt.prop_empty("volatile");
+                // ctsi,label always starts with the deterministic "_flashdrive<i>"
+                // (leading underscore is reserved for these internal names);
+                // the optional user label (if any) is appended after a comma.
+                std::string dt_label = "_flashdrive"s + std::to_string(i);
+                if (!f.label.empty()) {
+                    dt_label.append(",").append(f.label);
+                }
+                fdt.prop_string("ctsi,label", dt_label);
```
edubart: I would prefer to use `-`, since this appear in the file tree at `/proc/device-tree`, names with `,` is a bit awkward because you have to use `\,` when mentioning it in bash scripts, making life more difficult.

src/interpret.cpp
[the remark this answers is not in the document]
edubart: I mean, the code is fine. Copilot is confused.

src/interpret.cpp
```diff
@@ -5229,6 +5241,127 @@ static FORCE_INLINE execute_status execute_C_ADDW(const STATE_ACCESS a, uint64_t
...
 
+/// \brief Implementation of the C.LBU instruction (Zcb extension).
+template <typename STATE_ACCESS>
+static FORCE_INLINE execute_status execute_C_LBU(const STATE_ACCESS a, uint64_t &pc, uint64_t mcycle, uint32_t insn) {
+    [[maybe_unused]] auto note = dump_insn(a, pc, static_cast<uint16_t>(insn), "c.lbu");
+    const uint32_t rd = insn_get_CIW_CL_rd_CS_CA_rs2(insn);
+    const uint32_t rs1 = insn_get_CL_CS_CA_CB_rs1(insn);
+    const uint32_t uimm = insn_get_C_LS_B_uimm(insn);
+    return execute_C_L<uint8_t>(a, pc, mcycle, rd, rs1, uimm);
+}
```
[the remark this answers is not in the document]
edubart: We will need to add new riscv-tests suite, which require many changes, better to leave for my RVA23 PR to do this.

src/machine-config.cpp
```diff
@@ -32,7 +32,7 @@
 #include "os-filesystem.hpp"
 #include "riscv-constants.hpp"
 
-static constexpr uint32_t archive_version = 6;
+static constexpr uint32_t archive_version = 7;
```
edubart, line 35: Since we are bumping versions and breaking compatibility already, better update emulator version and MARCHID in a commit (in `src/Makefile`)

src/machine-config.hpp
```diff
@@ -77,24 +79,30 @@ struct ram_config final {
 
 /// \brief DTB state config
 struct dtb_config final {
-    std::string bootargs{
-        "quiet earlycon=sbi console=hvc0 root=/dev/pmem0 rw init=/usr/sbin/cartesi-init"}; ///< Bootargs to pass
-                                                                                           ///< to kernel
+    std::string bootargs{"quiet earlycon=sbi console=hvc0 root=/dev/pmem0 rw init=/usr/sbin/cartesi-init "
+                         "uio_pdrv_genirq.of_id=generic-uio"}; ///< Bootargs to pass to kernel
```
edubart: I would move `uio_pdrv_genirq.of_id=generic-uio` before `root=` so we can let `cartesi-machine.lua` remove ALL bootargs after `root=` when using `--no-root-flash-drive`.

src/cm-error.h
```diff
@@ -0,0 +1,47 @@
...
+    CM_ERROR_BAD_OPTIONAL_ACCESS = -14,
+    CM_ERROR_BAD_WEAK_PTR = -15,
+    CM_ERROR_BAD_FUNCTION_CALL = -16,
+    CM_ERROR_BAD_ALLOC = -17,
+    CM_ERROR_BAD_ARRAY_NEW_LENGTH = -18,
+    CM_ERROR_BAD_EXCEPTION = -19,
+    CM_ERROR_BAD_VARIANT_ACCESS = -20,
+    CM_ERROR_EXCEPTION = -21,
+    CM_ERROR_UNKNOWN = -22,
+} cm_error;
```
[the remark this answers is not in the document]
diegonehab: Should we care?

tests/dependencies
```diff
@@ -1,2 +1,2 @@
-https://github.com/cartesi/machine-guest-tools/releases/download/v0.17.2/rootfs-tools.ext2
-https://github.com/cartesi/machine-linux-image/releases/download/v0.20.0/linux-6.5.13-ctsi-1-v0.20.0.bin
+https://github.com/cartesi/machine-guest-tools/releases/download/v0.18.0-test1/rootfs-tools.ext2
+https://github.com/cartesi/machine-linux-image/releases/download/v0.21.0-test1/linux-6.5.13-ctsi-2-uio-test1-v0.21.0.bin
```
edubart, line 2: Should we merge this PR before or after making a release of linux-image/guest-tools?

tests/Makefile
```diff
@@ -54,8 +54,22 @@ LIBCARTESI_SO_JSONRPC=libcartesi_jsonrpc-$(MACHINE_EMULATOR_SO_VERSION).so
 NUM_JOBS ?= $(shell nproc)
 endif
 
+LUAROCKS_PATH := $(shell luarocks --lua-version 5.4 path --lr-path 2>/dev/null)
+LUAROCKS_CPATH := $(shell luarocks --lua-version 5.4 path --lr-cpath 2>/dev/null)
+export LUA_PATH_5_4 := $(LUA_PATH_5_4);$(LUAROCKS_PATH)
+export LUA_CPATH_5_4 := $(LUA_CPATH_5_4);$(LUAROCKS_CPATH)
```
edubart, line 60: LuaRocks revived 🫤 . With this change tests will become impossible to run anymore on systems without LuaRocks, even though they are not running coverage.

Why this was necessary? I think should be a problem of the user environment to deal with proper LuaRocks installation. Or can we move this line inside `coverage=yes` case so we can still run tests on machines without LuaRocks?

tests/misc/test-machine-c-jsonrpc-api.cpp
```diff
@@ -0,0 +1,722 @@
...
+    std::string host = address.substr(0, colon);
+    int port = std::stoi(address.substr(colon + 1));
+
+    int fd = ::socket(AF_INET, SOCK_STREAM, 0);
+    BOOST_REQUIRE_GE(fd, 0);
+
+    struct sockaddr_in sa{};
+    sa.sin_family = AF_INET;
+    sa.sin_port = htons(static_cast<uint16_t>(port));
+    ::inet_pton(AF_INET, host.c_str(), &sa.sin_addr);
```
edubart, line 87: This is a bit too low level, could just use Boost.ASIO instead.

tests/lua/spec-cm-cli.lua
```diff
@@ -0,0 +1,1592 @@
...
+            )
+        )
+        if pid then
+            os.execute("wait " .. pid .. " 2>/dev/null || kill " .. pid .. " 2>/dev/null")
+        end
+    end)
+end)
+
+-- Cleanup scratch directory (runs once when the module is loaded)
+os.execute("rm -rf " .. scratch)
```
edubart: `rm -rf <variable>` is scary, I would prefer just `rmdir` and ensure the code leaves the directory empty.

tests/lua/spec-cm-cli.lua
```diff
@@ -0,0 +1,1592 @@
...
+-- returned by machine:get_initial_config(), validating the full
+-- CLI -> machine builder -> get_initial_config() pipeline.
+
+local lester = require("cartesi.third-party.lester")
+local describe, it, expect = lester.describe, lester.it, lester.expect
+
+-- Scratch directory for temp files created by CLI invocations (stores, configs, etc.)
+-- Declared at module level so the post-describe cleanup can reference it.
+local scratch = os.tmpname() .. ".d"
+assert(os.execute("mkdir -p " .. scratch), "failed to create scratch directory")
```
edubart: You could use `utils.scope_exit()` (from `cartesi/utils.lua`) to auto delete the directory when scope exits  here.

tests/lua/spec-cm-cli.lua
```diff
@@ -0,0 +1,1592 @@
...
+                "#!/bin/sh\n%s %s --gdb=%s --max-mcycle=1000000 --no-init-splash --quiet >%s 2>&1 &\necho $!\n",
+                CLI_LUA,
+                CLI,
+                gdb_addr,
+                log
+            )
+        )
+        f:close()
+        os.execute("chmod +x " .. runner)
+        local pipe = io.popen(runner)
```
edubart, line 1748: Personally I don't likw the growing amount of use for `os.execute` and `io.popen` everywhere, we could use `luaposix` more for some basic operations like creating directories, changing permissions. I think assuming the host have these utilities a bit fragile.

tests/lua/spec-cm-cli.lua
```diff
@@ -0,0 +1,1592 @@
...
+            shquote(images_path),
+            CLI_LUA,
+            CLI,
+            table.concat(args, " "),
+            stdin_redir,
+            tmp_out,
+            tmp_err,
+            tmp_rc
+        )
+        os.execute(cmd)
```
edubart, line 122: What if instead of running with `os.execute`, we convert `cartesi-machine.lua` to a library we can require?

I wish we could abolish use of `os.execute` and `io.popen`, I personally refactored some tests to stop using them in the past.

src/cartesi-machine.lua
```diff
@@ -1541,37 +1640,38 @@ local options = {
...
-            flash_shared.root = nil
-            flash_create.root = nil
-            flash_truncate.root = nil
-            flash_read_only.root = nil
-            table.remove(flash_label_order, 1)
-            dtb.bootargs = dtb.bootargs:gsub(" root=$", "")
+            assert(flash_drives[1] and flash_drives[1].label == "root", "no root flash drive to remove")
+            flash_drives[1] = nil
+            flash_label_to_index.root = nil
+            dtb.bootargs = dtb.bootargs:gsub("^root=[^%s]+%s*", ""):gsub("%s+root=[^%s]+", "", 1)
```
edubart: This wasn't working before (my bad), but this option should also remove remove all `root=/dev/pmem0 rw init=/usr/sbin/cartesi-init` from bootargs, because all of them makes no sense without a root.

src/cartesi-machine.lua
```diff
@@ -183,22 +190,68 @@ where options are:
...
+        the machine always assigns the internal label "_nvramN" (where N is
+        the zero-based index). the user label is an optional additional name
+        that can be used to refer to the NVRAM in --replace-memory-range.
+        user labels must contain only alphanumeric characters, hyphens, and
+        underscores, must not start with an underscore (reserved for internal
+        use), must be unique across all flash drives and NVRAMs, and must be
+        at most 64 characters long.
+        if omitted, no user label is set.
+
+        start (required)
```
edubart: I think `start` should be optional, and auto set like we do with flash drives. Choosing a proper `start` is not easy for beginners. We have no manual on how to do it. Think people using this on Cartesi CLI.

src/cartesi-machine.lua
```diff
@@ -2095,67 +2196,54 @@ echo "
...
-            config.dtb.init = config.dtb.init .. cmd .. "\n"
+    end
+
+    for idx = 1, nvram_count do
+        local entry = nvrams[idx]
+        if entry then
+            local dt_label = "_nvram" .. #config.nvram
+            if not entry.length then entry.length = -1 end
+            config.nvram[#config.nvram + 1] = entry
+            config.dtb.init = config.dtb.init .. string.format("dev=$(nvram %s)\n", dt_label)
```
edubart, line 2247: Should we do any check for people who choose to use outdated guest-tools or linux-image?
Currently this happens:
```
$ cartesi-machine.lua --nvram=start:0x900000000,length:4096 --quiet

         .
        / \
      /    \
\---/---\  /----\
 \       X       \
  \----/  \---/---\
       \    / CARTESI
        \ /   MACHINE
         '

/usr/sbin/cartesi-init: /proc/device-tree/cartesi-machine/init: line 12: nvram: not found
chmod: : No such file or directory
Nothing to do.
```

I used both outdated kernel and rootfs here.

src/cartesi-machine.lua
```diff
@@ -2095,67 +2196,54 @@ echo "
...
+            if entry.read_only then
+                config.dtb.init = config.dtb.init .. 'chmod 0444 "$dev"\n'
```
edubart: Always prepend these cli utilities with `busybox ...`, eg `busybox chmod 0444 `.
Guest tools make no guarantees that any utility besides its own executable and busybox is available.

src/dtb.cpp
```diff
@@ -171,22 +171,51 @@ void dtb_init(const machine_config &c, unsigned char *dtb_start, uint64_t dtb_le
...
+            int i = 0; // NOLINT(misc-const-correctness)
+            for (const auto &n : c.nvram) {
+                fdt.begin_node_num("uio", n.start);
+                fdt.prop_string("compatible", "generic-uio");
+                fdt.prop_u64_list<2>("reg", {n.start, n.length});
+                std::string dt_label = "_nvram"s + std::to_string(i);
+                if (!n.label.empty()) {
+                    dt_label.append(",").append(n.label);
+                }
+                fdt.prop_string("ctsi,label", dt_label);
```
edubart: I am not convinced of the structure chosen here. I think the `cartesi-init` could be simplified by providing a better structure here.

Instead of creating `/run/cartesi/memoryranges/_nvram0` containing `/dev/uio0 0x0000000900000000 0x0000000000001000` during init, we could instead construct these values here, like:
```
/proc/device-tree/cartesi-machine/memory-ranges/_nvram0/label: userlabel
/proc/device-tree/cartesi-machine/memory-ranges/_nvram0/dev: uio0
/proc/device-tree/cartesi-machine/memory-ranges/_nvram0/start: 0x0000000900000000
/proc/device-tree/cartesi-machine/memory-ranges/_nvram0/length: 0x0000000000001000
/proc/device-tree/cartesi-machine/memory-range-labels/userlabel: _nvram0
```

Then `nvram` tool could just open `/proc/device-tree/cartesi-machine/labels/<label>` read it, and then open whatever it needs from `/proc/device-tree/cartesi-machine/`

Also `/run/cartesi/memoryranges/_nvram0` could just become symlinks to `/proc/device-tree/cartesi-machine/_nvram0`.

Splitting between files in separate files will reduce parsing logic and time. I think it will simplify `cartesi-init` and the memoryrange tool. The numbers don't need to be hexadecimal, could even be binary with use of `prop_u64`, even less parsing, no need for `fscanf`.

edubart (changes requested):

edubart:
I ran some benchmarks, using `nvram` is much faster than using `flash-drive`, and even faster than guest RAM itself. This is great! :smiley: 

```
Benchmark 1: cartesi-machine.lua --nvram=label:userlabel,start:0x900000000,length:64Mi --flash-drive=label:data,length:64Mi,mount:false,data_filename:mem,shared,mke2fs:false --volume=.:/mnt --user=root -- /mnt/mmap_stress --mem 64M --file /dev/pmem1 --rw-iters 1K --iters 128 --block-size 128
  Time (mean ± σ):      4.460 s ±  0.067 s    [User: 4.381 s, System: 0.064 s]
  Range (min … max):    4.405 s …  4.624 s    10 runs

Benchmark 2: cartesi-machine.lua --nvram=label:userlabel,start:0x900000000,length:64Mi,shared,data_filename:mem,shared --flash-drive=label:data,length:64Mi,mount:false,mke2fs:false --volume=.:/mnt --user=root -- /mnt/mmap_stress --mem 64M --file /dev/uio0 --rw-iters 1K --iters 128 --block-size 128
  Time (mean ± σ):     398.2 ms ±   4.7 ms    [User: 365.3 ms, System: 28.0 ms]
  Range (min … max):   391.0 ms … 403.8 ms    10 runs

Benchmark 3: cartesi-machine.lua --nvram=label:userlabel,start:0x900000000,length:64Mi,shared,data_filename:mem,shared --flash-drive=label:data,length:64Mi,mount:false,mke2fs:false --volume=.:/mnt --user=root -- /mnt/mmap_stress --mem 64M --rw-iters 1K --iters 128 --block-size 128
  Time (mean ± σ):     875.9 ms ±  13.8 ms    [User: 849.6 ms, System: 23.3 ms]
  Range (min … max):   855.7 ms … 891.7 ms    10 runs

Summary
  cartesi-machine.lua --nvram=label:userlabel,start:0x900000000,length:64Mi,shared,data_filename:mem,shared --flash-drive=label:data,length:64Mi,mount:false,mke2fs:false --volume=.:/mnt --user=root -- /mnt/mmap_stress --mem 64M --file /dev/uio0 --rw-iters 1K --iters 128 --block-size 128 ran
    2.20 ± 0.04 times faster than cartesi-machine.lua --nvram=label:userlabel,start:0x900000000,length:64Mi,shared,data_filename:mem,shared --flash-drive=label:data,length:64Mi,mount:false,mke2fs:false --volume=.:/mnt --user=root -- /mnt/mmap_stress --mem 64M --rw-iters 1K --iters 128 --block-size 128
   11.20 ± 0.21 times faster than cartesi-machine.lua --nvram=label:userlabel,start:0x900000000,length:64Mi --flash-drive=label:data,length:64Mi,mount:false,data_filename:mem,shared,mke2fs:false --volume=.:/mnt --user=root -- /mnt/mmap_stress --mem 64M --file /dev/pmem1 --rw-iters 1K --iters 128 --block-size 128
```

This is the script I asked Claude to create for me for testing this https://gist.github.com/edubart/cf149b0af834ff991415ac7b311abd70

src/cm-exception.hpp
diegonehab: system_error derives from runtime_error. It adds two fields that are not portable. Think a Windows client talking to a remote machine that is running on Linux. There is no way we could send the system_error from the server back through the wire and reconstruct in the client. I prefer for there not to be this difference in behavior between local and remote machines. So I downgrade system_error to runtime_error in the C api. Note that, before the recent changes, *all* remote exceptions were being transformed into runtime errors and all messages were different. Now, we can basically use the same test suit for local and remote machines. I actually think we should remove even more of the exception types from cm_error.

diegonehab: Even the what() field of system_error is not round-tripable.

src/cm-error.h
diegonehab: The reason was for cm-exception.hpp not to have to include the entire cm.h. But if you think this is not worth it, I can revert this change.

src/cm.h
diegonehab: Well, then it would have been called something else. :) It is the opposite of get_reg_address for addresses that are in fact regs. But it has to be a static function in machine, so it can't know about labels. Shall I just improve the documentation or change the name to something more obvious?

src/dtb.cpp
diegonehab: I remember seeing a constant somewhere... How about FW_RESV? :)

src/dtb.cpp
diegonehab: You mean instead of ctsi-label of ctsi,label? The "common practice" is to use commas, right? If you mean in the *value*, nobody would use commas there: it's just a separator that creates aliases for the same address range.

src/cartesi-machine.lua
diegonehab: This is a bit brittle, right? Ideally, we would not do gsubs to remove things from the existing bootargs. We should probably keep this out of the bootargs and only add in the end?

src/cartesi-machine.lua
diegonehab: I don't see a use case where the user creates one of these and then doesn't need to know where it landed. Even Solidity will have to know where this is, most likely. Unlike flash drives, where nobody cares, really. So I am not sold on the auto-starts...

src/cartesi-machine.lua
diegonehab: You had to work hard to get this error here, and IMO you deserve it. :) If you omit the nvram option, does it work? It should, right? Because `flashdrive` is there. I actually thought about this.

src/dtb.cpp
diegonehab: I think your proposal is significantly more complicated than having a single file per drive with all the info there.  There is no overhead that you can measure in any barely realistic situation. The memoryrange/nvram/flashdrive program already does all the required logic. The init is complicated only because getting the data from the device tree is complicated. If we want to simplify this, we could simply add a dedicated container at the root node, "/ctsi" or something, to the device tree and put the data there in a much more convenient form.

edubart: > I think your proposal is significantly more complicated than having a single file per drive with all the info there. 

Could be a single entry if want to simplify the tools.

>  The init is complicated only because getting the data from the device tree is complicated

It could be much simpler if the data it needs was already wells structured in DTB in a way that is simple here

> If we want to simplify this, we could simply add a dedicated container at the root node, "/ctsi" or something

That is what I meant here, we already do this, for instance both `/proc/device-tree/cartesi-machine/init` and `/proc/device-tree/cartesi-machine/entrypoint` is constructed here in DTB file. Both automatically appears as files and are read by `cartesi-init`.

src/cartesi-machine.lua
edubart: > Even Solidity will have to know where this is, most likely

The use cases for nvram is not only for stuff that needs its address to be known at Solidity.

> I don't see a use case where the user creates one of these and then doesn't need to know where it landed

People could still lookup by checking memory ranges layout, the problem would switch to just lookup what was decided, instead having to decide one.

For example I was just trying to benchmark, and I had to decide an address to create a nvram to run the benchmark, but I didn't care about the address, I ended using random numbers in the address until it worked, because I didn't remember by mind what ranges are addressable and I was too lazy to calculate an address by hand.

src/cartesi-machine.lua
edubart: The point of doing gsubs was to have a single source of default bootargs. Also I don't want to remove `root=/dev/pmem0 rw init=/usr/sbin/cartesi-init` from default bootargs, because then we will have to explain in every C API use case example what bootargs are and what you should do with them. Currently all C API examples can completely avoid even mentioning bootargs.

src/dtb.cpp
edubart: I meant `ctsi-label`, both using `,` and `-` are common if you lookup for other DTB fields. But think ones using `-` are easier and cleaner to work with from bash scripts.

src/cm-error.h
edubart: I prefer that this change is reverted, to make life easier who generate bindings from the C API for libcartesi, by having to parse just `cm.h` and have all types it needs self contained.

src/dtb.cpp
diegonehab: Working on it. I think I can simplify this a lot.

src/cm-exception.hpp
edubart: But I read that some other exceptions are still being downgraded and restructured to runtime_error, like std::regex_error:

```c
        case CM_ERROR_RUNTIME_ERROR:
        case CM_ERROR_REGEX_ERROR: // here
        case CM_ERROR_EXCEPTION:
        case CM_ERROR_UNKNOWN:
        default:
            throw std::runtime_error(message);
```

Then also remove `CM_ERROR_REGEX_ERROR`, or make both downgrade to runtime_error

src/cm-error.h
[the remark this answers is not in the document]
edubart: There is already one comment from me about this.

src/cm.h
edubart: I mean an example on the documentation, or make clear in the documentation this is not the label. I think the function name is fine.

src/dtb.cpp
edubart: `FW_RESV` looks like to short and could conflict with other open source projects, `AR_FW_RESV_LENGTH`  ?

src/cartesi-machine.lua
diegonehab: We don't have to mention it, but we have to know in Lua what was put there by C++, right? And then remove it somehow. It is brittle. Let's find a cleaner way.

src/dtb.cpp
diegonehab: This is just in dtb.cpp.

src/dtb.cpp
diegonehab: I removed the whole thing anyway. Take a look.

src/dtb.cpp
diegonehab: Check.

edubart: Much better IMO :smile:

src/cartesi-machine.lua
diegonehab: Check now.

src/cartesi-machine.lua
diegonehab: Check now.

diegonehab:
Alright. I’m done if you are.

edubart (approved):

Issue #388, opened 2026-06-10
Better handling of revert root hash


diegonehab:
## Context

  In the rollups flow, when the guest application rejects an advance-state input, the
  canonical machine state must go back to what it was before the input was fed. The host
  implements this by rolling back to a snapshot. Dispute resolution, however, must be able
  to *prove* this transition, and a verifier (EVM or ZK) cannot reconstruct the pre-input
  state on its own. The machine state hash it can compute after replaying the rejecting
  run is the post-reject hash, not the pre-input hash.

  The solution has two sides.

  On the write side, the pre-input root hash must already live inside the hashed state by
  the time the rejection happens. `machine:send_cmio_response()` now takes the revert root
  hash as its first, logged, verifiable argument and commits it into the `revert_root_hash`
  leaf of the shadow state (at `AR_SHADOW_REVERT_ROOT_HASH_START`) as part of its proof,
  replacing the old unverified two-step where the host poked the hash separately via
  `write_revert_root_hash()`.

  On the read side, the post-operation verifiers apply one rule. After replaying the
  operation, if the machine ended in a manual yield with reason rx-rejected, the expected
  root hash after is the stored revert root hash. Otherwise it is the machine's actual
  root hash. The rule lands in exactly two operations. `reset_uarch` ends the
  meta-instruction in the uarch dispute, and `step` ends the bisected mcycle range in the
  ZK dispute. `verify_step_uarch` never substitutes (intermediate uarch hashes are real
  machine hashes), and `verify_send_cmio_response` never substitutes (the send clears
  `iflags.Y`, so its post-state can never be rejected).

  The substitution makes bisection consistent because it switches the canonical timeline
  from the rejecting run onto the snapshot exactly once. Before the yield, commitments
  carry real hashes from the rejecting run. At the yield, the substituted hash. After the
  yield, the canonical state *is* the pre-input state S, whose pending accepted yield makes
  it a fixed point of the step, so the tail is constant and provable from the snapshot.
  One precondition is essential and unverifiable by any verifier. The committed revert
  root hash must be that of a machine in a manual-yield-accepted state. The dispute
  protocol must bind it to the agreed pre-input boundary hash, which in the rollups flow
  always satisfies this (the machine boots into an rx-accepted yield).

  ## Possible solutions

  The adopted design folds the rejection check and reversion inside the operations
  themselves, so verifiers and the dispute contracts need no substitution logic.

[...]

PR #389, opened 2026-06-10
Prove revert-on-reject state transitions

diegonehab:
Part of #388 (the issue stays open for the Dave integration items). The companion PR regenerates the Solidity side: cartesi/machine-solidity-step#98

  ## Problem

  When the guest rejects an advance-state input, the canonical machine state goes back to
  the pre-input state, but dispute verifiers could not prove that transition. The
  pre-input root hash was poked into the machine by the host through a separate,
  unverified `write_revert_root_hash()` call, and no verifier ever substituted it for the
  machine root hash after a rejection.

  ## Solution

  The first commit implements the write side. `send_cmio_response`,
  `log_send_cmio_response`, and `verify_send_cmio_response` take the revert root hash as
  their first argument, threaded as `const_machine_hash_view` through every interface
  layer (machine, i_machine, local and jsonrpc machines, the JSON-RPC wire protocol, the
  C API, and the Lua bindings) and as a `bytes32` alias in the Solidity-translatable
  core. The core commits the hash into the `revert_root_hash` shadow leaf as a single
  full-leaf logged write right after the `iflags.Y` precondition check, so the value
  becomes part of the proof. The record/replay rx-buffer write logic is hoisted into
  label-parameterized helpers shared with the new accessor.

  The second commit implements the read side. `uarch_reset_state` resets the uarch as
  before, then reads `iflags.Y`, conditionally reads `htif.tohost`, decodes the
  dev/cmd/reason fields, and on a rejected manual yield calls the new `revert_state()`
  accessor, which reads the revert leaf (one logged access carrying the value) and
  switches the canonical root hash atomically. The replay accessor replaces its context
  root hash, so `verify_reset_uarch` is untouched. The record accessor reports the
  resulting hash through a caller-owned context, so `log_reset_uarch` needs no branch.
  For the ZK path, `replay_step_state_access::finish()` applies the same rule, reading
  the revert leaf from the shadow registers page that every step log already contains,
  and `log_step` records the substituted hash in the log header. The risc0 guest inherits
  the behavior with no source changes. Unknown yield reasons (GIO domains, accepted,
  exception) cause no substitution and no error. The physical machine is never reverted,
  a rejected `log_reset_uarch` still resets the uarch.

  cm.h gains the HTIF tohost/fromhost field shifts and masks, exposed through the Lua
  module so the machine-solidity-step constants generator can query them.

  ## Tests

  - machine-bind (local and jsonrpc) covers the new access shapes, the rejected reset and
    step round trips against the revert hash, tampered revert leaf values, and the
    accepted-reason control case.
  - The corrupted-log specs cover the new revert-hash write access of the send.
  - The uarch json log generator emits a rejected-reset fixture consumed by the
    machine-solidity-step Foundry tests.
  - Verified locally: all Lua suites, C API tests, the 54 uarch tests, uarch.bin
    cross-build, and the risc0 cpp guest cross-build.


mpernambuco (dismissed):
LGTM. I like that the reject revert is now intrinsic to reset instead of synthesized on the solidity side

mpernambuco (dismissed):

src/machine.cpp
mpernambuco, line 1700: since this overwrites the shadow tlb without going through write_??verified_tlb, could a pending write-tlb dirty page be lost here? should it mark_write_tlb_dirty_pages() before the memcpy?

mpernambuco (dismissed):
LGTM  Nice call with  `i_accept_dirty_pages`

src/machine.cpp
diegonehab: You are right, if you call this function when the machine already has some hot entries. I'll fix it.

diegonehab: Fixed.

mpernambuco (dismissed):

mpernambuco (dismissed):

mpernambuco (approved):

PR #394, opened 2026-07-14
feat!: model cycle overflows as persistent fixed points

diegonehab:
## Problem

  Machine and microarchitecture cycle limits were not handled consistently across execution, logging, verification, and
  hash collection.

  In particular, reaching a cycle limit needed to produce a fixed point without requiring an additional machine-state
  transition. Calls targeting the current cycle also always reported that they had reached the target, even when the
  current state already implied overflow, halt, or yield.

  ## Solution

  This PR models machine and microarchitecture cycle overflow as derived, state-preserving fixed points.

  - Machine overflow is derived from `mcycle` and `imcyclemax`.
  - Microarchitecture overflow is derived from `uarch_cycle` and `UARCH_CYCLE_MAX`.
  - A machine at either limit does not advance further.
  - Overflow takes precedence over halt, manual yield, and reaching the target cycle.
  - Calls targeting the current cycle return the break reason implied by the current state.
  - Hash collectors return the same break reason that the corresponding run operation would return.
  - `imcyclemax` is updated when an accepted advance-state input starts a new processing interval.
  - Rollup limits are centralized under `ROLLUP_*` constants.

  The microarchitecture step now reads `uarch_cycle` before checking `uarch_halt`. After executing an instruction, it
  checks overflow immediately after incrementing the cycle, before reporting a halt produced by that instruction. This
  keeps the emulator, recorded access logs, replay verification, and Solidity step verifier aligned.

  ## API changes

  This PR adds explicit machine and microarchitecture overflow break reasons across the C++, C, Lua, and JSON-RPC APIs.

  Notable renames include:

  - `uarch_halt_flag` to `uarch_halt`
  - configuration field `halt_flag` to `halt`
  - `reached_target_cycle` to `reached_target_uarch_cycle`
  - `cycle_overflow` to `uarch_cycle_overflow`
  - `CM_CMIO_LOG2_MAX_OUTPUT_COUNT` to `CM_ROLLUP_LOG2_MAX_OUTPUT_COUNT`

  The obsolete halt-value constants used to distinguish ordinary halt from cycle overflow were removed.

  ## Testing

  The updated tests cover:

  - Machine and microarchitecture overflow fixed points
  - Overflow precedence over halt and manual yield
  - Calls targeting the current cycle
  - Logging and verification at cycle limits
  - Machine-cycle and microarchitecture-cycle hash collection
  - Local and remote machine APIs
  - C, Lua, and JSON-RPC break-reason mappings
  - Microarchitecture access-log ordering


mpernambuco (dismissed):
LGTM

edubart (dismissed):

src/cartesi-machine.lua
```diff
@@ -2902,172 +3092,779 @@ end
...
+        flush_pending_outputs(m, advance)
+        claim:end_epoch()
+    elseif is_mcycle_overflow(break_reason) then
+        report_mcycle_overflow(m)
+        flush_pending_outputs(m, advance)
+        claim:end_epoch()
+    elseif is_yielded_manual(break_reason) then
+        save_cmio_output_proofs(advance)
+        claim:end_epoch()
     end
```
mpernambuco, line 3859: what if an input uses its whole budget without yielding?
should there be an elseif is_target_mcycle(break_reason) here?

diegonehab: In that case we can’t complete the computation hash, right?

diegonehab: You mean because of max mcycle passed in the command line? Because the overflow condition is checked there.

src/cartesi-machine.lua
```diff
@@ -2902,172 +3111,748 @@ end
...
+                claim:end_epoch()
+                return
             else
+                -- An unexpected manual yield is still a fixed point. Finalize the claim before
+                -- reporting the protocol error so callers can dispute the computation that led
+                -- to it. In particular, the uarch claim may still need to pad a selected period
+                -- that execution never reached.
+                claim:end_epoch()
                 error("unexpected manual yield reason")
             end
```
mpernambuco, line 3842: what  will the  backup_closer do after this for ends with active backup_machine, revert? should we commit()?

diegonehab: Very nice catch. In these weird error conditions, the computation hash is perfectly defined: they are sticky fixed points that will never change (send_cmio_response is a noop on them). This is true of an unexpected manual yield, an mcycle overflow, halt, or exception.  We should commit. Will do.

diegonehab: Implemented.

mpernambuco (approved):

PR #398, opened 2026-07-28
feat: add a stored-machine revert mode to cartesi-machine

diegonehab:
Add machine::rename_stored, threaded through cm.h/cm.cpp, i-machine, local-machine, jsonrpc-machine, the machine.rename_stored jsonrpc method, and the Lua machine:rename_stored binding. The rename refuses to replace an existing destination atomically (renameat2 RENAME_NOREPLACE on Linux, renamex_np RENAME_EXCL on macOS, with a checked-rename fallback), and syncs the affected parent directories so the rename is durable. remove_stored now syncs its parent directory too.

The jsonrpc handlers for clone_stored, remove_stored, and sync_stored no longer require a machine loaded in the session, since none of them act on one.

cartesi-machine gains --revert-mode=<fork|stored|none>, replacing --no-revert (kept as an alias for none). Mode stored implements cmio snapshot/commit/revert over the stored directory of a machine loaded with sharing:all or created with --create, cloning it to <dir>.revert, and works both locally and against a remote jsonrpc server. The new sync key of --load flushes the loaded backing stores to permanent storage before exit. --dump-constants prints the constants of the cartesi Lua module as shell assignments, for use by scripts that need to name HTIF yield reasons and similar values. Bash completion now completes enum positional values for compound options like --revert-mode.

The one-input-per-invocation shell-loop section of the docs is replaced by a single stored-mode invocation, and the run-rolling-calculator-from-storage.lua recipe mirrors the CLI stored mode exactly (destroy, clone, sync, in-place load, rename_stored on rollback).

Add spec-backing-stores rename tests (including no-overwrite), C API rename_stored tests, spec-cm-cli tests for revert-mode validation and a full stored revert flow against a remote server, and a test pinning that a guest halt during --cmio-advance-state surfaces the halt payload in the exit status.


src/jsonrpc-discover.json
```diff
@@ -256,7 +256,7 @@
     },
     {
       "name": "machine.remove_stored",
-      "summary": "Removes a stored machine instance from a directory",
+      "summary": "Removes a stored machine instance from a directory and makes the removal durable",
```
mpolitzer, line 259: Does "makes the removal durable" mean will force a sync to disk after the operation? I find the wording a bit confusing as is.

My 2c is to omit the second part (leave as was originally), or reword into something like " and force a sync to disk".

mpolitzer (dismissed):
No support for windows on these APIs then

tuler (approved):

src/jsonrpc-discover.json
diegonehab: "Durable" is a technical term, right?

diegonehab: Maybe change to persistent?

mpolitzer (approved):

edubart (approved):

